PR #120, opened 2025-05-27
fix: scrapbook language ID overlapped with other languages from other extensions

tnaum-ms:


tnaum-ms:
We've not updated any of the `scopeName`s as we plan a larger follow-up work on scrapbook support.
This PR focuses on resolving the language server id overlap with other extensions.

PR #185, opened 2025-06-27
 Password Prompt Fix

nishthaAhujaa:
When the connection string uses username:<password>, we want the password prompt to come. Hence, treating <password> as empty password

### DIFF
--- a/src/vscodeUriHandler.ts
+++ b/src/vscodeUriHandler.ts
@@ -77,10 +77,7 @@ async function handleConnectionStringRequest(
 
     // Parse the connection string
     const parsedCS = new ConnectionString(params.connectionString!);
-    if (
-        parsedCS.password &&
-        decodeURIComponent(parsedCS.password.trim()) === '<password>'
-    ) {
+    if (parsedCS.password && decodeURIComponent(parsedCS.password.trim()) === '<password>') {
         parsedCS.password = '';
     }
 


src/vscodeUriHandler.ts
```diff
@@ -77,6 +77,9 @@ async function handleConnectionStringRequest(
 
     // Parse the connection string
     const parsedCS = new ConnectionString(params.connectionString!);
+    if (parsedCS.password && decodeURIComponent(parsedCS.password.trim()) === '<password>') {
+        parsedCS.password = '';
```
xingfan-git, line 81: In line 161, the parsed connection string (parsedCS) is stored in a storage item, and I suppose this allows the connection to be used after reopening the VSCode window. I'm unsure if setting the password as an empty string would affect the current experience. Could you check whether your change might cause the connection to become invalid after reopening VSCode please?

tnaum-ms (changes requested):
@nishthaAhujaa thanks for the contribution. I get the idea, but I’d kindly ask you to reconsider this change.

`<password>` is unusual, but still a valid value. Using `<password>` it as a signal feels risky, since someone might actually use it as their password. The current way of leaving out the password field from the connection string already gives the right behavior.

A better long-term fix would be to add a show/hide toggle to the input UI. We both know folks on extension teams, so we could reach out, ask for support, upvotes, or ideas. This issue would be a good starting point: https://github.com/microsoft/vscode/issues/221397

Thanks again. I’d love to see this move forward in a more solid direction.

PR #226, opened 2025-08-19
Use estimatedDocumentCount to Get Rough Document Count

xingfan-git:


tnaum-ms (approved):
🚀 Perfect, thank you!

Please review what happens with our task progress reporting in case an unexpected value comes back from the `estimatedDocumentCount`, like a `0`.

I'm not sure if we have safeguards in place where we compute the progress.

PR #330, opened 2025-10-31
Sync `next` with `main` after two patch releases

tnaum-ms:


tnaum-ms:
branch protections are making this PR difficult to handle.

PR #411, opened 2025-12-03
Add telemetry for LLM relates calls

xingfan-git:


src/services/promptTemplateService.ts
```diff
@@ -61,7 +62,18 @@ export class PromptTemplateService {
                     }),
                 );
             } catch (error) {
-                // Log error and fall back to built-in template
+                // Log telemetry for template load operation (automatically includes result='Failed' when error is thrown)
+                void callWithTelemetryAndErrorHandling(
```
tnaum-ms: @xingfan-git FYI: attempt to create entries for "what is happening", not to say something has failed. This gives us better data, and still logs failures (`success == false`)

PR #412, opened 2025-12-04
fix: Discovery View always started with the RU provider added

tnaum-ms:
This pull request refactors the logic for promoting discovery providers in the `DiscoveryBranchDataProvider` class to ensure promotions are only shown to users who have already engaged with discovery providers. The method responsible for this logic is now public for testing purposes, and the checks for when to show a promotion have been clarified and improved.

### Improvements to promotion logic

* Changed the promotion eligibility check to use active discovery provider IDs (`activeDiscoveryProviderIds`) instead of all registered providers, ensuring promotions are only shown to users who have already interacted with discovery providers.
* Added detailed documentation to the `addDiscoveryProviderPromotionIfNeeded` method, clarifying its logic and making it public for easier testing.

### Codebase simplification

* Removed redundant reading of `activeProviderIds` before the promotion logic, streamlining the code and reducing unnecessary state reads.


tnaum-ms:
👍

PR #454, opened 2026-01-20
Feat: Add URL-encoding detection and retry logic for cluster authentication

### LINKED ISSUE #444: Improve: Warn users about an URL Encoded password
**Issue:**
Users may attempt to connect using a password that is URL-encoded, which results in failed authentication and a poor experience.

**Proposed Solution:**
On connection failure, analyze the password locally. If it appears to be URL-encoded, notify the user and suggest decoding it. Optionally, we could offer to decode the password automatically.

vaibh123540:
This PR addresses an issue where users often provide URL-encoded passwords leading to authentication failures. The logic waits for a connection failure after which:

1. It checks for URL-encoding a password using the ```decodeURIComponent```method and regex search using ```/%[0-9A-Fa-f]{2}/```.
2. If it finds URL-encoding in the password, it suggests the user about it and offers a "retry with decoded password" option.
3. If the user chooses to retry, the connection request is made again with decoded password.
4. In case of failure user is prompted with the regular error message (no option for URL-encoding), in case of success the connection is established and the decoded password is updated in permanent storage so that the same error doesn't appear again in the next session.

Files Changed:
```DocumentDBClusterItem.ts```: Updated ```authenticateAndConnect``` catch block to include the described logic after a connection failure.

Testing performed:
I manually tested the workflow for an URL encoded password and a non URL encoded password. I tested both for a successful connection and an unsuccessful connection in the "retry with decoded password" logic.

The overall logic seems to have improved the User Experience.


tnaum-ms:
Thank you for the contribution! 🙏 The original PR was closed because it targeted a branch that no longer exists (`rel/release-0.7.1`), but your work is valid and appreciated.

The review is continuing in #594 where maintainer adjustments have been applied before merging to `next`:
- Resolved the merge conflict with the updated `next` branch
- Fixed a type-safety issue in error handling (`(error as Error).message` → proper `instanceof` guard)
- Fixed cache key to use `clusterId` instead of `this.id` (required for stability when connections are moved between folders)

PR #545, opened 2026-04-02
feat: add Copy Reference context menu for databases and collections

bgaeddert:
Adds a "Copy Reference" right-click option to database and collection nodes in the sidebar. Database copies the db name, collection copies db.collection dot notation.

PR #587, opened 2026-04-16
feat: add Copy Reference context menu for databases and collections [reviewed]

tnaum-ms:
Reviewing and merging: https://github.com/microsoft/vscode-documentdb/pull/545

---

## Summary

Thanks to @bgaeddert for kicking this off and for the inspiration! 🎉

Starting from the original contribution that added "Copy Reference" for databases and collections, we've expanded and refined the feature:

### What changed

- **Unified into a single command** — One `copyReference` command handles all node types, detecting the node via `instanceof` at runtime. No more separate commands per node type.
- **Extended to Index nodes** — "Copy Reference…" now appears on database, collection, and index context menus.
- **QuickPick format picker** — Instead of copying a single fixed format, the user picks from contextual options:
  - **Database**: Name, Shell Command (`use dbName`), Qualified Name (`host/dbName`)
  - **Collection**: Name, Namespace (`db.coll`), Shell Reference (`db.coll`), Shell Command (`db.getCollection("coll")`)
  - **Index**: Name, Key Definition (`{ field: 1 }`), Shell Command (`db.coll.getIndexes().find(...)`)
- **Smart escaping** — Names with special characters (spaces, parens, etc.) automatically use the `db.getCollection("...")` form. Dot-notation options are hidden when they would produce broken syntax.
- **Consistent menu placement** — "Copy Reference…" always appears in its own group just above the Refresh group, across all node types.
- **Uses `context.ui.showQuickPick`** with `suppressPersistence: true` for stable ordering — follows the same pattern as other wizard-style commands in the codebase.

### Feedback requested

@bgaeddert — the main change from your original PR is the addition of a QuickPick step to choose the format. We'd love your feedback on whether that extra step is acceptable to you. If you'd prefer a zero-click experience, we could follow up with a PR that adds a user setting to define a preferred default format — when set, it would copy directly without showing the picker.


tnaum-ms:
@bgaeddert 

first-contribution

sajeetharan (approved):

PR #593, opened 2026-04-21
[on hold until 0.8.0 ships] feat: add sorting functionality to DataViewPanelTable and update Quer…

bgaeddert:
Serverside Column Sorting on header click

- Add column-header sort support to the collection view data table: clicking a column header applies a server-side sort for that field, and clicking again cycles through ascending → descending → none (tri-state).
- Sorting updates the Sort input in the QueryEditor so users can see (and edit) the applied sort document.

### DIFF
--- a/src/webviews/documentdb/collectionView/collectionViewContext.ts
+++ b/src/webviews/documentdb/collectionView/collectionViewContext.ts
@@ -98,6 +98,7 @@ export type CollectionViewContextType = {
             limit: number;
         };
         setJsonSchema(schema: object): Promise<void>; //monacoEditor.languages.json.DiagnosticsOptions, but we don't want to import monacoEditor here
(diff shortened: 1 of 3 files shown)


tnaum-ms:
@bgaeddert we'll be revisiting the table component once 0.8.0 ships (0.8.0 is a massive one, that's why we're a bit slower atm). I'll keep this PR on hold until the table component is updated.

tnaum-ms:
@bgaeddert in case you've got time, take a look this pre-build: https://github.com/microsoft/vscode-documentdb/discussions/595

tnaum-ms:
@bgaeddert Hi, this PR was not forgotten. We recently shipped the Interactive Shell feature with 0.8.0, and are wrapping up some reviews. I'll have the table/tree components revisited then and we'll come back here.

PR #617, opened 2026-04-27
feat(kubernetes): complete DocumentDB discovery workflow

### LINKED ISSUE #581: Feature: Add Kubernetes Multi-Cloud Service Discovery Plugin
### Background

DocumentDB for VS Code includes an open Service Discovery architecture that allows plugins to authenticate with a vendor environment, navigate its resource hierarchy, discover database services, and surface connection strings back to the extension. Today, the available plugins target Azure-hosted environments:

- **Azure Cosmos DB for MongoDB (RU)**
- **Azure DocumentDB (vCore)**
- **Azure VMs (DocumentDB)**

A fourth plugin for **MongoDB Atlas** discovery is in progress (see #531).

This issue proposes a new **Kubernetes Service Discovery** plugin so users running DocumentDB-compatible databases on Kubernetes — regardless of cloud or cluster provider — can discover and connect to them without manually copying connection strings or service endpoints. The plugin works with any reachable Kubernetes cluster: AKS, EKS, GKE, OpenShift, on-premises, local (minikube, kind, k3s), and any other conformant distribution.

### Prior Work

There is no Kubernetes-specific code in the extension today. The Kubernetes API access layer does not exist yet and is out of scope for this issue. It will need to be introduced as part of this work, relying on an **external** Kubernetes client library (e.g., `@kubernetes/client-node`) rather than anything built into the extension.

This issue therefore covers:

1. Selecting and integrating an appropriate external Kubernetes client package.
2. Building the Discovery View integration, user-facing wizards, and tree view experience on top of it.

> ⚠️ **Dependency footprint is a first-class concern.** See [Extension Size & Memory Impact](#extension-size--memory-impact) below.

### Goal

Users should be able to:

1. **Enable the Kubernetes plugin** from the Service Discovery panel, the same way they enable existing providers.
2. **Authenticate** via the **Manage Credentials** entry point. The user selects one or more kubeconfig contexts (local file or pasted kubeconfig YAML), each representing a cluster. Credentials are stored securely in VS Code's secret storage.
3. **Browse their Kubernetes estate** in the Discovery tree view — Contexts (clusters), Namespaces, and individual `Service` resources that expose a DocumentDB-compatible port — with cluster provider, namespace, and service type (ClusterIP / LoadBalancer / NodePort) visible at a glance.
4. **Filter visible resources** via the **Filter** entry point to narrow which contexts and/or namespaces appear, mirroring how Azure plugins filter by tenant and subscription.
5. **Select a service** to have its connection endpoint resolved and added to the Connections view.

### UX Expectations

- Follow the same UX patterns established by existing plugins — use the existing service discovery plugin infrastructure, including the `DiscoveryProvider` registration, the **Manage Credentials** entry point, and the **Filter** entry point.
- **Credential management wizard** should let users:
  - **Enable or disable** kubeconfig contexts from `~/.kube/config` or a custom path — rather than importing or copying credentials, the wizard could simply present all contexts discovered in the user's kubeconfig file(s) and let them toggle which ones the extension should monitor. This is worth exploring as the primary flow since kubeconfig is already the standard Kubernetes credential store; the UX question is how to surface the concept of "showing a context in DocumentDB" without asking users to paste or re-enter anything.
  - Name or alias each enabled context for display in the tree (the raw kubeconfig context name is often opaque).
  - Remove or disable previously enabled contexts.
  - Unlike Azure plugins (which delegate to the VS Code Azure account), this plugin surfaces the user's existing kubeconfig material — the UX should make that clear and guide users who do not yet have a kubeconfig set up.
- **Filter wizard** should let users scope which kubeconfig contexts and namespaces are visible in the tree, similar to how Azure plugins filter by tenant / subscription.
- Tree items should have the same affordances as existing plugins: refresh, retry on error, learn more link.
- See **Managing Azure Discovery** ([docs/user-manual/managing-azure-discovery.md](../../user-manual/managing-azure-discovery.md)) for the credential and filter UX to use as inspiration..

### Proposed Tree Structure

```
Kubernetes Discovery (root)
└── context: my-aks-cluster (AKS / eastus)
    ├── namespace: default
    │   ├── svc: documentdb-primary  [LoadBalancer :27017]
    │   └── svc: documentdb-replica  [ClusterIP :27017]
    └── namespace: staging
        └── svc: documentdb-staging  [NodePort :30017]
└── context: local-kind (kind / localhost)
    └── namespace: dev
        └── svc: mongo-dev  [ClusterIP :27017]
```

### Connection String Resolution

The exact strategy for each service type, and the port-forward lifecycle, should be confirmed during implementation (see "Exploration Needed" below).

### Exploration Needed

The following questions should be investigated as part of this work:

- **Kubernetes client library selection** — Evaluate `@kubernetes/client-node` and any lighter alternatives. Consider bundle size, tree-shakeability, in-cluster vs. out-of-cluster support, and maintenance status.
- **Service discovery heuristics** — How should the plugin decide which `Service` objects are DocumentDB-compatible? Options: user-defined label selectors, or an annotation convention (e.g., `documentdb.vscode.extension/discovery: "true"`)?
- **Multi-kubeconfig sources** — Beyond `~/.kube/config`, should the plugin support in-cluster service accounts (for users running VS Code inside a pod), or KUBECONFIG environment variable paths? What is the priority order?
- **Port-forward lifecycle** — If a port-forward tunnel is opened for a ClusterIP service, when should it be torn down? On disconnection, on extension deactivation, or on explicit user action?
- **Credential secret convention** — Is there a de-facto standard annotation or label for pointing a Service at a credential Secret? Should the plugin support multiple conventions (e.g., Helm chart defaults) or require an explicit annotation?
- **Cluster identification** — kubeconfig contexts do not always carry human-readable cloud-provider metadata. What information is feasible to surface (provider type, region) without making extra API calls?
- **Minimum RBAC** — What Kubernetes RBAC permissions are required for read-only discovery? The plugin should document the minimum `ClusterRole` or `Role` + `RoleBinding` users need to create, and ideally link to a copy-pasteable manifest.

### Extension Size & Memory Impact

Adding a Kubernetes client library is a **non-trivial dependency**. Unlike the Azure plugins — which reuse `@azure/*` libraries already bundled in the extension — a Kubernetes client would be new weight added to every install, even for users who never use Kubernetes discovery.

This issue includes a soft requirement to **measure and evaluate** the impact:

- Profile the compiled VSIX size before and after adding the Kubernetes client dependency.
- Measure VS Code extension host memory at idle and during active Kubernetes discovery.
- Identify whether the client (or large parts of it) can be lazily loaded — i.e., only `require()`'d when the Kubernetes plugin is first activated by the user.

**Expected outcome:** either a validated lightweight architecture (e.g., lazy `import()` at plugin activation time, with acceptable overhead), or a **follow-up issue** that introduces a general _dynamic plugin activation_ mechanism for all service discovery providers — allowing plugins to load their dependencies on demand rather than at extension startup. This would benefit the Atlas plugin and any future provider that brings its own SDK.

### Out of Scope

- Building a general dynamic plugin activation mechanism (may become a follow-up).
- Kubernetes cluster management features beyond discovery (scaling, deployments, secrets management).
- Changes to the core discovery plugin interface.
- Support for non-Kubernetes container runtimes (Docker Compose, Podman) — those may be addressed by a separate plugin.

### References

- **Existing plugin UX patterns:** `src/plugins/service-azure-vm/` and `src/plugins/api-shared/azure/`
- **Credential management infrastructure:** `src/plugins/api-shared/azure/credentialsManagement/`
- **Filter infrastructure:** `src/plugins/api-shared/azure/subscriptionFiltering/`
- **MongoDB Atlas discovery issue** (sister issue for non-Azure non-Kubernetes cloud): #531
- **Service Discovery overview:** `docs/user-manual/service-discovery.md`
- **Managing Azure Discovery** (UX to emulate): `docs/user-manual/managing-azure-discovery.md`
- [Kubernetes API reference — Service](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/service-v1/)
- [Kubernetes RBAC documentation](https://kubernetes.io/docs/reference/access-authn-authz/rbac/)

guanzhousongmicrosoft:
## Summary

Fixes #581.

This PR completes Kubernetes multi-cloud service discovery for DocumentDB-compatible targets and wires it into the existing Service Discovery provider architecture.

### Implemented feature

- Adds a Kubernetes discovery provider that can be enabled from the Service Discovery panel, plus New Connection > Service Discovery > Kubernetes support.
- Integrates `@kubernetes/client-node` as the external Kubernetes client requested by the issue.
- Supports kubeconfig sources for the Kubernetes client's default loading behavior (including `KUBECONFIG` and `~/.kube/config`), a custom kubeconfig file, or pasted kubeconfig YAML.
- Validates the selected kubeconfig source before saving setup changes, surfaces the discovered context count, and fails inside the setup wizard for invalid or zero-context kubeconfigs.
- Uses source-only credential setup for UX parity with Azure DocumentDB discovery: adding/reconfiguring Kubernetes asks only for the kubeconfig source, enables all contexts by default, and does not prompt for aliases, context selection, or namespace selection.
- Adds a context-only Filter action that starts with the currently visible contexts selected, so users can deselect contexts and reopen the filter without losing state.
- Implements lazy tree browsing for contexts, namespaces, and services so slow or unreachable contexts remain visible and failures appear as retry/error nodes instead of blocking root expansion.
- Shows namespaces containing DocumentDB targets as expandable and sorts them first, while namespaces without DocumentDB targets remain visible as non-expandable informational leaves.
- Discovers DKO `documentdb.io/preview` `dbs` resources before generic Kubernetes Service fallback.
- Supports generic service discovery by explicit opt-in annotation/label or known DocumentDB API-compatible TCP ports.
- Resolves LoadBalancer, NodePort, ExternalName, and ClusterIP endpoints; ClusterIP targets use managed local port-forward tunnels.
- Persists non-secret Kubernetes port-forward metadata on saved connections so ClusterIP tunnels can be restarted when a saved Kubernetes connection is reopened.
- Adds duplicate-connection handling that compares parsed hosts and Kubernetes port-forward identity without printing raw credential-bearing connection strings.
- Adds provider docs, RBAC guidance, local test-cluster setup/teardown scripts, a detailed handoff plan, and targeted Jest coverage across provider activation, credential setup, filtering, tree browsing, endpoint resolution, port-forwarding, and connection flows.

### Issue scope decisions

- The issue originally suggested Manage Credentials could enable/disable contexts and collect aliases. Based on UX testing, this PR intentionally keeps Manage Credentials source-only and uses Filter for context visibility, matching Azure DocumentDB discovery behavior more closely.
- Namespace filtering and namespace selection prompts were removed. The tree still shows every namespace, but only namespaces with DocumentDB targets expand.
- Cluster provider metadata is best-effort and derived from kubeconfig context/cluster/server values; the implementation does not make extra cloud-provider API calls.
- The PR documents minimum RBAC permissions for namespaces, services, nodes, endpoints, pods/portforward, secrets, and DKO resources in `docs/user-manual/service-discovery-kubernetes.md`.

### Dependency footprint

- `@kubernetes/client-node` is added as the Kubernetes API client.
- Kubernetes runtime code uses dynamic `import('@kubernetes/client-node')` paths and the extension webpack config preserves dynamic imports so Kubernetes client code is loaded on demand rather than during extension startup.
- Root tree browsing is intentionally lazy and does not scan namespaces/services until the user expands a context or starts the Kubernetes connection wizard.

### Review follow-up fixes

- Hardened ClusterIP port-forward lifecycle cleanup: successful kubeconfig changes and provider removal stop tunnels, while cancelled/failed credential setup preserves existing tunnels.
- Port-forward stop/dispose now invalidates pending starts, closes active TCP sockets/WebSocket streams, destroys error streams, and prevents stale pending starts from registering after provider state changes.
- Managed Kubernetes port-forward tunnels no longer offer "Use existing" when a different Kubernetes target already owns the requested local port; users must choose a different local port to avoid saving a connection to the wrong target.
- ClusterIP backend resolution now carries the selected Service port name into tunnel metadata and matches Endpoints ports by name, with single-port fallback and no arbitrary multi-port selection.
- Generic Kubernetes Service targets now preserve the selected Service port name, including remapped multi-port ClusterIP services such as `port: 27017`, `targetPort: 10260`, `name: documentdb`.
- DKO-discovered targets keep `tlsAllowInvalidCertificates=true` by default because `status.tls.ready` confirms certificate availability, not that the workstation trusts the issuing CA.
- Default kubeconfig loading now rejects the `@kubernetes/client-node` synthetic `loaded-context` / `http://localhost:8080` fallback as missing kubeconfig, without rejecting explicit custom/inline localhost kubeconfigs.
- Namespace service pre-scan keeps the current "DocumentDB namespaces first" sort but uses bounded concurrency to avoid fanning out requests across all namespaces at once.
- Kubernetes target leaves retain the standard `treeItem_documentdbcluster` context value so Create database, Copy connection string, and Open Interactive Shell remain available alongside Kubernetes-specific actions.
- Kubernetes root empty/error states now show guided recovery actions: Configure kubeconfig, Manage Filter when filters hide every context, Open Kubernetes discovery docs for kubeconfig issues, and Retry.

## Secret handling

- Default and custom-file kubeconfig sources are not copied into extension storage. The extension stores only the selected source type and, for custom files, the kubeconfig file path in global state.
- Pasted kubeconfig YAML is stored only in VS Code SecretStorage (`kubernetes-discovery.inlineKubeconfig`). It is not stored in globalState, logs, telemetry, docs, or connection properties; it is deleted when the user switches back to a non-inline kubeconfig source.
- Kubernetes credential discovery reads credentials from Kubernetes Secrets only when needed for a selected target. DKO admin secrets and annotated generic service secrets are resolved from the target namespace, decoded in memory, and used to create the connection; secret values are not written to output logs or telemetry.
- Saved connections continue to use the extension's existing secret/native-auth storage path for username/password data. The additional Kubernetes properties persisted by this PR are non-secret port-forward metadata: context, namespace, service name, remote port, local port, and local endpoint details required to restart a tunnel.
- Connection strings generated from discovery are credential-free. Existing redaction paths are preserved for display/logging, and duplicate detection compares parsed hosts plus native-auth identity instead of printing raw credential-bearing connection strings.

## Validation

- `npm run l10n`
- `npm run prettier-fix`
- Targeted Jest suites for Kubernetes discovery, credential setup, filtering, tree nodes, discovery wizard, port-forwarding, Add Discovery Registry, retry authentication, open collection view, collection tree, and command classification
- `npm run lint`
- `npm run build`
- `git --no-pager diff --check`


guanzhousongmicrosoft:
Superseded by #621.

PR #621, opened 2026-04-29
feat(kubernetes): add multi-source service discovery

### LINKED ISSUE #581: Feature: Add Kubernetes Multi-Cloud Service Discovery Plugin

guanzhousongmicrosoft:
## Summary

Adds first-class **Kubernetes service discovery** to the DocumentDB VS Code extension: users can connect to DocumentDB clusters running in Kubernetes (DKO operator-managed or generic DocumentDB-API services) without manually building connection strings or running `kubectl port-forward`.

Supersedes #617.

### Discovery flow
- New **Kubernetes** entry under the Discovery view (visible by default — see below).
- **Multi-source kubeconfig** management: combine the default kubeconfig, additional file sources, and pasted (inline) YAML sources persisted via VS Code Secret Storage.
- **Drag-and-drop** kubeconfig files onto the Discovery view to register them as sources.
- **Per-context display aliases** (rename a context without modifying YAML or saved connections).
- **DocumentDB Kubernetes Operator (DKO)** awareness: operator-managed clusters are surfaced first, generic Services with DocumentDB-compatible ports as fallback.

### Connectivity
- **Port-forward tunnel manager** (`portForwardTunnel.ts`) with:
  - Race-safe lifecycle (generation-counter invalidation re-checked at every async boundary — retry sleep, post-bind, post-`showWarningMessage`, pre-register).
  - EADDRINUSE single-retry with explicit "Use existing" prompt.
  - EACCES treated as a hard config error (not collision) — explicit recovery hint, never offers a fake localhost connection.
  - Reachability classification per service type (LoadBalancer-direct / NodePort-routed / ClusterIP-port-forward-required) with icon + description + tooltip.
- **Saved connection auto-restart**: `ensureKubernetesPortForward` lives in `DocumentDBClusterItem.beforeCachedClientConnect`, so re-opening a saved K8s cluster transparently re-establishes the tunnel.
- **Read-only copy**: copying a ClusterIP connection string no longer starts a tunnel; copied strings carry metadata so the saved-connection warning toast can fire on later use.

### Settings (scoped subset of #20)
- `documentDB.serviceDiscovery.kubernetes.portForward.localPortStrategy` — `matchRemote` (default) or `autoSelect`.
- `documentDB.serviceDiscovery.kubernetes.portForward.localPortBase` — default `27100`.
- `autoSelect` uses bind-attempt scan on `127.0.0.1`, capped at 100 candidates, falls back to the remote service port.

### Cross-cutting changes that landed alongside
- **Discovery providers visible by default** (`discoveryProviderVisibility.ts`): inverted from a "what to show" allowlist to a hidden-providers denylist, with one-time migration from the legacy `activeDiscoveryProviderIds` and the `azure-discovery` → `azure-mongo-vcore-discovery` rename normalization. Unknown provider ids are preserved in storage so plugin registration timing changes never silently drop a user's hide preference.
- **`BaseExtendedTreeDataProvider.failedChildrenCache`** — error/recovery nodes are cached so Refresh returns the same recovery tree; only explicit Retry/Reload/mutating commands clear the cache via `resetNodeErrorState`.
- **`DocumentDBShellPty.writeLine`** normalizes bare `\n` to `\r\n` so multi-line error output no longer staircases in the pseudoterminal.
- **`webpack.config.ext.js`** no longer externalizes `socks` — the transitive `@kubernetes/client-node` dep is now bundled, fixing the TDZ at K8s import time.

## Coverage — bug-bash repo `tnaum-ms/vscode-documentdb-bugbash-090`

All **30 closed issues** were verified through this branch:

| Verdict | Count | Issues |
|---|---|---|
| ✅ Resolved | 23 | #1, #3, #5, #7, #8, #9, #10, #11, #13, #14, #15, #16, #17, #18, #22, #24, #25, #29, plus #2, #4, #12, #19 (with minor caveats) |
| ⚠️ Resolved with caveats | 4 | #6 (deferred Reachability later added by #21), #20 (4-of-5 settings deferred — documented), #21 (kubectl snippet deferred), #23 (KUBECONFIG-merge intentionally out of scope) |
| 🟡 Likely resolved (pending repro) | 1 | #30 (no repro from OP; fix is sound) |
| ⚪ Wontfix-justified | 2 | #27, #28 (deferred to a planned table-component update) |

Verification was done in two full passes: an initial 30-issue × 5-model review (150 verdicts), then a corrected re-verification on the 6 issues that warranted a second look (30 more verdicts). All findings are recorded.

## Review history

This PR went through two rounds of independent multi-agent code review (5 models per round: gpt-5.4, gpt-5.5, claude-opus-4.6, 4.7, 4.8).

**Round 1** surfaced 15 findings against the pre-rebase state. All 11 worth fixing pre-merge were addressed and committed:

| Fix | Subject |
|---|---|
| `c12412f5` | `portForwardTunnel`: leak-free `errStream` cleanup + explicit EACCES error |
| `dd950d1f` | `kubernetesClient`: IPv6 brackets, multi-entry KUBECONFIG, `createCoreApi` mutation doc |
| `cfc4b24b` | `sourceStore` / `aliasStore`: rename-order preservation, inline-add race protection, ensureCache invalidate-during-load race, alias write serialization |
| `15311649` | K8s `treeId` aligned with `clusterId` suffix; `removeKubeconfigSource` resets the failed-children cache for the removed node; `discoveryProviderVisibility` preserves unknown ids |

**Round 2** (run against the rebased branch): 4 of 5 reviewers completed (gpt-5.4 was cancelled before starting); the 4 completed reviewers verified every round-1 fix at HEAD and returned **unanimous READY-TO-MERGE / Approve** with zero blockers and zero should-fix. Findings explicitly deferred (filed as future-issue candidates rather than blockers): inline-add cosmetic ordering between unrelated adds, kubectl-share snippet (#21), exec-auth trust prompt for pasted/dropped kubeconfigs (matches kubectl industry norm), per-reconnect storage-read perf hint.

## Validation

All five required steps pass cleanly at HEAD:

```text
npm run l10n          # bundle regenerated (1462 keys after rebase)
npm run prettier-fix  # no changes
npm run lint          # clean
npx jest --no-coverage   # 129 suites, 2404 / 2404 tests
npm run build         # clean
```

## Testing matrix

| Area | Mechanism |
|---|---|
| `sourceStore` + `aliasStore` race semantics | targeted Jest with `Promise.allSettled × 3` concurrent-add cases + globalState fault injection |
| `portForwardTunnel` lifecycle, cancellation, EADDRINUSE retry, EACCES error | targeted Jest with fake-timer cancellation sweeps |
| `KubernetesServiceItem` reachability tooltip + copy semantics | parameterized Jest over all 6 reachability states |
| K8s `treeId` ↔ `clusterId` alignment | `findClusterNodeByClusterId` lookup test |
| Cross-platform kubeconfig paths (`%USERPROFILE%`, KUBECONFIG splitting) | `path.sep`-tolerant Jest assertions |
| Discovery visibility denylist + migration | 4-path migration coverage incl. `azure-discovery` rename |

Not exercised in CI (manual smoke required for full sign-off): live `kubectl` against a real cluster, SOCKS-proxied connection through the newly-bundled `socks` module.

## Known intentional non-goals

- `KUBECONFIG`-style **path-list merging** (kubectl's full merge semantics). We follow the operationally-equivalent "first existing entry wins" + `loadFromDefault` fallback. Matches `@kubernetes/client-node` behavior, not full kubectl semantics.
- `kubectl port-forward …` **share-snippet** from #21 — metadata is now persisted to enable this as a follow-up enhancement.
- Four of the five **settings proposed in #20** (`namespaceScanConcurrency`, `additionalPorts`, `showEmptyNamespaces`, DKO CRD version escape hatch) — deferred to separate design issues.
- Table-component **theme/contrast polish** (#27) and **double-click row toggle** (#28) — both wontfix on the bug-bash branch, tied to a planned table-component refresh.
- **Exec-auth trust prompt** for pasted/dropped kubeconfigs — matches kubectl industry norm; can be added later if policy demands explicit per-source trust.


tnaum-ms:
**Plugin boundary: custom inline "+" command**

Location: `package.json` `contributes.menus` + `KubernetesRootItem.ts`

The Kubernetes plugin introduced a custom `discoveryView.kubernetes.addSource` command and wired it as an inline `$(add)` button directly on its root tree node. This steps outside the established plugin API boundary.

The correct approach is to **extend the discovery plugin API** with a first-class `"add source"` entry point that any plugin can implement, rather than each plugin rolling its own inline command. This keeps the contract between the extension host and plugins clean and prevents future plugins from doing the same ad-hoc.

Note: "add a kubeconfig source" is fundamentally what a user is mentally doing. It is closer to **"add connection"** semantics than credential management, so the right API to introduce is something like `onAddSource`, not repurposing `manageCredentials`.

tnaum-ms:
@guanzhousongmicrosoft we'll have to carefully review the PR on the code level. I noticed that some changes made on `next` have been removed by commits in this PR. Did you experience a larger merge conflict?

tnaum-ms:
> @guanzhousongmicrosoft we'll have to carefully review the PR on the code level. I noticed that some changes made on `next` have been removed by commits in this PR. Did you experience a larger merge conflict?

My bad, I didn't pull, forget it :)

tnaum-ms:
UX review is now in progress. Working from the bug-bash-090 catch-up document at `docs/ai-and-plans/PRs/621-kubernetes-discovery/bugbash-090-kubernetes-ux-review.md`.

tnaum-ms:
### Pre-merge review fix 2/3 — surface disabled TLS validation (review finding 2.1)

Commit `499ef0aa` — `fix(kubernetes): surface disabled TLS validation on discovered targets`

The insecure default is **accepted** (the DKO gateway typically serves a self-signed cert, so strict verification would break the common path), but it is no longer silent:

- **Tooltip:** discovered target nodes now show a `⚠️ Security: TLS/SSL certificate validation disabled` line in their hover tooltip — the same treatment the Connections-view node gives an emulator with security disabled. It reads from the resolved connection params (`tlsAllowInvalidCertificates=true`) rather than assuming, so it stays accurate if the param set ever changes. The warning is tooltip-only (not the always-visible description), to avoid cluttering the row.
- **Docs:** new **"Connection security (TLS/SSL)"** section in the user manual explains the default and why it exists, and notes that **after saving** a discovered target you can edit the connection and remove `tlsAllowInvalidCertificates=true` to re-enable certificate validation (e.g. when the server presents a trusted CA-signed cert) — phrased consistently with the existing local-connection TLS guidance.

Gating on the DKO `tlsReady` signal remains a possible future refinement and is intentionally **not** done here. Marked finding **2.1** as ✅ done in the review doc.

tnaum-ms:
### Pre-merge review fix 3/3 — document kubeconfig `exec`-plugin trust model (review finding 2.3)

Commit `d024d9bb` — `docs(kubernetes): warn that kubeconfig exec plugins run local binaries`

A kubeconfig can reference [`exec` credential plugins](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins) (the standard mechanism used by AKS/EKS/GKE — `kubelogin`, `aws`, `gke-gcloud-auth-plugin`) that run on the user's machine when a source is expanded. Added a prominent trust-model warning to the **"Add a kubeconfig source"** section of the user manual, advising users to only add, paste, or drop kubeconfigs they trust — the same trust model as `kubectl`. No code change.

This commit also reconciles the pre-merge review doc (doc-gap table, summary table, recommendation) to reflect 2.1/2.3/2.7 as resolved. Marked finding **2.3** as ✅ done.

---

**Remaining:** finding **2.2 (no request timeout/cancellation on K8s API calls)** is still under investigation — checking whether `@kubernetes/client-node` already applies a built-in request timeout before deciding on any code change.

tnaum-ms:
### Pre-merge review fix 1/3 — pin `@kubernetes/client-node` (review finding 2.7)

Commit `f9b42633` — `build(deps): pin @kubernetes/client-node to 1.4.0`

Removed the caret range (`^1.4.0` → `1.4.0`) in `package.json` and synced `package-lock.json`. A future `npm install` can no longer silently pull a newer 1.x with behavior changes in the exact areas Kubernetes discovery depends on (`onInvalidEntry`, the `loadFromDefault` synthetic-config shape, `exec` auth), which the code currently hardens against defensively.

Marked finding **2.7** as ✅ done in the pre-merge review doc.

tnaum-ms (dismissed):

sajeetharan (dismissed):
LGTM

sajeetharan (approved):
LGTM

PR #650, opened 2026-05-15
docs: clarify getMongoClient caller

### LINKED ISSUE #643: Revisit: getMongoClient() JSDoc is misleading; only used by main thread scanCollectionSchema
## Context

## Suggested solutions

## Definition of done

MukundaKatta:
## Summary
- update getMongoClient() JSDoc to describe its current extension-host schema-scan usage
- remove the misleading query playground evaluator / ServiceProvider wording

Fixes #643.

## Testing
- git diff --check

### DIFF
--- a/src/documentdb/ClustersClient.ts
+++ b/src/documentdb/ClustersClient.ts
@@ -341,8 +341,8 @@ export class ClustersClient {
 
     /**
      * Returns the underlying MongoClient instance.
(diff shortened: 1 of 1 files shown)


tnaum-ms:
@MukundaKatta Thank you for your contribution, I'll close this one as we're about the simplify the ClusterClient class around the `0.9.*` release.

PR #670, opened 2026-05-21
fix: display shard key in collection tooltip

### LINKED ISSUE #661: Show shard key in collection tooltip 🔑
## Problem

Sharded collections in the tree view don't display their shard key anywhere visible. Users must use external tools (Azure Portal, shell) to check which field(s) a collection is sharded on. This information is important for understanding query performance and data distribution.

The raw MongoDB `listCollections()` response includes an `options` field that can contain shard key information, but the current `CollectionItemModel` interface only preserves `name` and `type`, discarding this data.

## Expected Behavior

When a collection has a shard key, it should be displayed in the collection's tooltip:

```
### my-collection

`Collection`

---

**Database:** my-database
**Documents:** 5.2k
**Shard Key:** { userId: "hashed" }
```

## Development Hints

Since this is a `good first issue`, here are some hints on where to start:

**Step 1: Extend the model** in `src/documentdb/ClustersClient.ts`

The `CollectionItemModel` interface currently only captures `name` and `type`:

```typescript
export interface CollectionItemModel {
    name: string;
    type?: string;
}
```

Add an optional `options` field (or specifically a `shardKey` field) to preserve the shard key from the MongoDB response. The raw `listCollections()` response already contains this data, but it gets discarded during the type assignment:

```typescript
const rawCollections = await this._mongoClient.db(databaseName).listCollections().toArray();
const collections: CollectionItemModel[] = rawCollections; // options field is lost here
```

**Step 2: Update the tooltip** in `src/tree/documentdb/CollectionItem.ts`

In `buildTooltip()`, add a section for the shard key when it exists. Format the key object as a readable string (e.g., `{ userId: "hashed" }` or `{ region: 1, date: 1 }`).

**Notes:**
- Not all collections are sharded. The shard key section should only appear when the data is present.
- The shard key is a key-value object where values indicate the sharding strategy (e.g., `1` for range, `"hashed"` for hash-based).
- The `listCollections()` response format may vary. Test with both sharded and non-sharded collections.
- Some DocumentDB configurations may not return shard key info at all. Handle missing data gracefully.

## Files to Modify

| File | Change |
|------|--------|
| `src/documentdb/ClustersClient.ts` | Extend `CollectionItemModel` to include shard key data |
| `src/tree/documentdb/CollectionItem.ts` | Add shard key section to `buildTooltip()` |

## Acceptance Criteria

- [ ] `CollectionItemModel` preserves shard key data from the `listCollections()` response
- [ ] Sharded collections show the shard key in the tooltip
- [ ] Non-sharded collections are unaffected (no empty shard key section)
- [ ] Shard key is formatted readably (e.g., `{ field: 1 }` or `{ field: "hashed" }`)
- [ ] No changes to tree item label, description, or icon

lte-z:
Fixes #661

## Changes

This PR exposes shard key information for sharded collections in the tree view tooltip.

- Extended `CollectionItemModel` to include optional `shardKey` field
- Extracted shard key from MongoDB `listCollections()` response in `ClustersClient`
- Rendered shard key in `CollectionItem` tooltip when present
- Added localization entry for "Shard Key"

## Behavior

- Sharded collections now display their shard key in tooltip, e.g.:

```
Database: my-database
Documents: 5.2k
Shard Key: { userId: "hashed" }
```

- Non-sharded collections are unaffected (no shard key section shown)

## Notes

- Handles missing or undefined shard key safely
- Supports both numeric (`1`) and hashed shard key formats


lte-z:
@microsoft-github-policy-service agree

tnaum-ms:
@lte-z Thank you for your contribution and for addressing Copilot review. I'll review your PR tomorrow.

tnaum-ms:
I pushed one follow-up commit on top of yours.

**What changed:** The shard key entries are now each wrapped in their own inline code span instead of being joined into a single one. So a two-key shard key renders as:

`userId: 1`, `tenantId: "hashed"`

**Why:** The original code applied `escapeMarkdown` to key and value strings before placing them inside a backtick code span. Inside a code span the Markdown renderer never interprets special characters, so the escape backslashes were rendered literally (e.g. `user\_id` showed up as `user\_id` with a visible backslash). Wrapping each entry in its own code span removes the need for escaping entirely and also makes multi-key shard keys easier to scan.

tnaum-ms:
Hey @lte-z, thanks for not just one, but the three PRs (#670, #674, #679) focused and quality contributions.

first-contribution

I'll have to work on the badges I use :)

tnaum-ms (approved):

lte-z:
Thank you so much for the warm welcome and for merging this!

I also appreciate the follow-up commit and the explanation about the inline code spans. That makes sense to me now, especially for shard keys with underscores or multiple fields.

And thank you for the kind words about the three PRs. I’m really glad they were helpful, and I’d be happy to keep contributing to vscode-documentdb 🙂

PR #683, opened 2026-05-27
Update getMongoClient JSDoc

### LINKED ISSUE #643: Revisit: getMongoClient() JSDoc is misleading; only used by main thread scanCollectionSchema

leakans:
Fixes #643

## Summary

Updates the `getMongoClient()` JSDoc to clarify that the current caller is the extension-host schema scan path, not the playground worker/evaluator.

This PR intentionally keeps the change documentation-only and does not alter the public API surface.

## Validation

- npm run prettier-fix
- npm run lint
- npm run build


tnaum-ms:
Hi @leakans -- thank you for picking this up and for the clean, focused fix!

After reviewing, we've decided to close this PR as **won't fix** for now. Here's the context:

Issue #643 was filed by me as a personal planning note and hadn't been triaged yet -- it was effectively in a "do not start" draft state. We should have labeled it more clearly to avoid exactly this kind of confusion, and I'm sorry about that. We'll be improving our workflow to make it obvious when an issue is still in draft and not ready for contributions.

The reason we're not merging even though the fix is technically correct: we're planning a broader refactoring of the `ClustersClient` / `MongoClient` surface in the near future, which will likely remove or replace `getMongoClient()` with a narrower API. The JSDoc would be overwritten as part of that work, making this change short-lived.

Again, thank you for the contribution and for running through the full validation checklist -- that's exactly the kind of diligence we appreciate. We hope this doesn't discourage you from contributing in the future. Keep an eye on our issues labeled [`good first issue`](https://github.com/microsoft/vscode-documentdb/labels/good%20first%20issue) for well-scoped, ready-to-work items.

PR #692, opened 2026-05-29
feat(tree): show index count on Indexes folder node

hanhan761:
@
## Summary
- Adds lazy-loaded index count as the `description` on `IndexesItem` tree items (e.g., "5 indexes" or "1 index")
- Follows the same pattern used by `CollectionItem` for document counts (fire-and-forget background fetch with stale-check guard)
- Count includes both regular indexes and search indexes when supported by the platform

## Issue
Fixes #659

## Changes
| File | Change |
|------|--------|
| `src/tree/documentdb/IndexesItem.ts` | Add `indexCount`, `isLoadingCount`, `isCurrent` fields; add `loadIndexCount()` / `fetchAndUpdateCount()` methods; show count in `getTreeItem()` description |
| `src/tree/documentdb/CollectionItem.ts` | Call `loadIndexCount()` on `IndexesItem` after creation in `getChildren()` |
| `l10n/bundle.l10n.json` | Extract new strings "1 index" and "{0} indexes" |

## Verification
- [x] `npm run build` — TypeScript compilation passes across all packages
- [x] `npm run lint` — ESLint passes with no errors
- [x] `npx jest` — 100 test suites, 1984 tests all passing
- [x] `npm run prettier-fix` — formatting confirmed
- [x] `npm run l10n` — localization bundle regenerated
@

PR #695, opened 2026-05-29
fix(tree): address Copilot review findings for index count feature

hanhan761:
@
## Summary
Addresses the Copilot review findings on PR #692 (index count on Indexes folder node). Three targeted fixes plus a cache optimization:

### Bug fixes
1. **Retry on failure**: `loadIndexCount()` now uses `typeof` check so a prior failed fetch (`null`) is retried after re-expand (was: `!== undefined` which treated null as "loaded")
2. **Loading flag leak**: `fetchAndUpdateCount()` wraps `isLoadingCount` reset in try/finally guarded by `isCurrent()`, so early bailouts never leave the instance permanently stuck
3. **Stale-check wiring**: `CollectionItem.getChildren()` now passes `isCurrent` predicate through to `IndexesItem`, properly discarding background work when the collection is refreshed/re-expanded

### Cache optimization
4. **Avoid duplicate listIndexes**: Background count fetch now caches the combined index list; `getChildren()` reuses it instead of making a second round-trip

## Related
- PR #692 (original feature)
- Copilot review: https://github.com/microsoft/vscode-documentdb/pull/692
@


hanhan761:
Superseded — the fix commit has been cherry-picked onto the original PR branch (feat-659-index-count) and is now part of #692.

PR #696, opened 2026-05-29
docs: fix misleading getMongoClient() JSDoc

### LINKED ISSUE #643: Revisit: getMongoClient() JSDoc is misleading; only used by main thread scanCollectionSchema

hanhan761:
@
## Summary
Fix the `getMongoClient()` JSDoc which incorrectly stated the method is used by the query playground evaluator. The playground worker runs in its own `worker_thread` with a separate `MongoClient` instance. The sole caller is actually `scanCollectionSchema` on the extension host main thread.

## Issue
Fixes #643
@

### DIFF
--- a/src/documentdb/ClustersClient.ts
+++ b/src/documentdb/ClustersClient.ts
@@ -342,8 +342,11 @@ export class ClustersClient {
 
     /**
      * Returns the underlying MongoClient instance.
-     * Used by the query playground evaluator to create a `@mongosh` ServiceProvider
-     * that reuses the existing, authenticated connection.
+     *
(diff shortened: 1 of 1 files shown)


tnaum-ms:
Thank you for this contribution! We apologize for the delayed review, we had the v0.9.0 release in the pipeline and it took up a lot of our capacity. We are closing this PR because the JSDoc for getMongoClient() was updated on main as part of recent internal work, so this change is already covered. We appreciate you identifying the misleading documentation and taking the time to address it!

PR #697, opened 2026-05-29
docs: fix getMongoClient() JSDoc to reflect actual caller

### LINKED ISSUE #643: Revisit: getMongoClient() JSDoc is misleading; only used by main thread scanCollectionSchema

hanhan761:
@
## Summary
- Updated the `getMongoClient()` JSDoc to accurately describe its usage
- Removed the misleading claim that it's used by the query playground evaluator
- Documented that the playground worker runs in a separate thread with its own MongoClient

## Issue
Fixes #643

## Changes
| File | Change |
|------|--------|
| `src/documentdb/ClustersClient.ts` | Fix `getMongoClient()` JSDoc |

## Verification
- JSDoc-only change — no runtime impact
- Verified the only caller is `scanCollectionSchema.ts` (extension host / main thread)
- Verified `playgroundWorker.ts` creates its own `MongoClient` in a separate worker thread

@


tnaum-ms:
Thank you for this contribution! We apologize for the delayed review, we had the v0.9.0 release in the pipeline and it took up a lot of our capacity. We are closing this PR because the JSDoc fix it contains is already on main, and the additional changes it bundles (concurrency limiter, index count updates, menu conditions) are each covered by separate, more focused PRs that are easier to review and track individually. We appreciate your thoroughness!

PR #698, opened 2026-05-29
ci: unify concurrency group to avoid double CI runs on release branches

### LINKED ISSUE #666: CI: double-run on `release/**` branches when forward-merge PR is open
## Problem

Every push to a `release/X.Y` branch triggers CI **twice** when a forward-merge PR (`release/X.Y → main`) is open:

| Event | Trigger | Concurrency group |
|---|---|---|
| `push` | `release/**` in push trigger | `github.run_id` (unique — never cancelled) |
| `pull_request` | PR targets `main` | `github.head_ref` = `release/X.Y` |

The two concurrency groups never cancel each other, so both runs complete in full. This wastes Actions minutes during every release stabilization cycle.

## Why it was accepted for now

This is the same behaviour that existed before with `next` + the `next → main` PR. Fixing it is non-trivial (any concurrency key that merges the two groups has side-effects on other flows), so it was deferred.

## Options to revisit

1. **Remove `release/**` from the `push` trigger** — CI then only runs via PR events. Build-size cache won't be seeded for direct release-branch pushes (small loss; cache falls back to the last `main` entry).
2. **Smarter concurrency key** — `${{ github.event.pull_request.number || github.ref }}` makes push runs on the same ref cancel each other, but still doesn't merge the push/PR groups.
3. **Separate leaner workflow for release branches** — triggered only on `push` to `release/**`, skipping the full PR reporting jobs.

hanhan761:
## Summary
When a forward-merge PR (`release/X.Y -> main`) is open, each push to the release branch triggered CI **twice** -- once via `push` and once via `pull_request`. The two runs had different concurrency groups and never cancelled each other, doubling Actions minutes.

**Fix**: Change the concurrency group from `github.head_ref || github.run_id` to `github.head_ref || github.ref_name`.

Both the push to `release/X.Y` and the PR from `release/X.Y -> main` now resolve to the same group (`release/X.Y`) and cancel each other. `release/**` stays in the push trigger, preserving CI coverage for direct release-branch pushes.

### Before
```yaml
concurrency:
    group: ${{ github.head_ref || github.run_id }}
```

### After
```yaml
concurrency:
    group: ${{ github.head_ref || github.ref_name }}
```

## Issue
Fixes #666

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,7 +34,11 @@ on:
             - feature/**
 
 concurrency:
-    group: ${{ github.head_ref || github.run_id }}
+    # Use head_ref for PRs, ref_name for pushes. When a forward-merge PR
+    # (release/X.Y → main) is open, both the push to release/X.Y and the
+    # PR trigger resolve to the same group (release/X.Y) so they cancel
+    # each other instead of running CI twice.
+    group: ${{ github.head_ref || github.ref_name }}
     cancel-in-progress: true
 
 jobs:


tnaum-ms:
Thank you again for this contribution. We are genuinely amazed by both the quality and the volume of improvements you have submitted. We also apologize for the delayed review while the 0.9.0 release was in flight. This PR is focused, clear, and very helpful for CI reliability.

tnaum-ms (approved):
Thanks again for the excellent contribution. Approved after rebase and full validation pass.

PR #694, opened 2026-05-29
perf(azure): throttle parallel tenant sign-in checks with shared concurrency limiter

hanhan761:
## Summary
Replace unbounded `Promise.all` fan-outs against the Azure Identity endpoint with a shared `ConcurrencyLimiter` (concurrency: 5).

For corporate users with many tenants, the original code issued unlimited parallel `isSignedIn` calls against Microsoft Entra -- the same anti-pattern fixed for `estimateDocumentCount` in #685.

## Issue
Fixes #688

## Changes
| File | Change |
|------|--------|
| `src/plugins/api-shared/azure/tenantSignInLimiter.ts` | **New** -- shared `LimitedRunner` (concurrency: 5) wrapping `createConcurrencyLimiter` |
| `src/plugins/api-shared/azure/credentialsManagement/SelectAccountStep.ts` | Import limiter, wrap `isSignedIn` calls |
| `src/plugins/api-shared/azure/subscriptionFiltering/InitializeFilteringStep.ts` | Import limiter, wrap `isSignedIn` calls |

## Verification
- [x] `npm run build` -- TypeScript compilation passes
- [x] `npm run lint` -- ESLint passes
- [x] `npm run prettier-fix` -- formatting confirmed

Issue #688, opened 2026-05-28
perf: throttle parallel Azure tenant sign-in checks


tnaum-ms:
## Context

Surfaced during the review for #685.

## Location

Two near-identical fan-outs against the Azure Identity endpoint:

- [src/plugins/api-shared/azure/credentialsManagement/SelectAccountStep.ts](https://github.com/microsoft/vscode-documentdb/blob/main/src/plugins/api-shared/azure/credentialsManagement/SelectAccountStep.ts) around line 146
- [src/plugins/api-shared/azure/subscriptionFiltering/InitializeFilteringStep.ts](https://github.com/microsoft/vscode-documentdb/blob/main/src/plugins/api-shared/azure/subscriptionFiltering/InitializeFilteringStep.ts) around line 67

Both do:

```ts
await Promise.all(tenants.map(async (tenant) => isSignedIn(tenant.tenantId, tenant.account)));
```

## The shape of the problem

For most users (1-10 tenants), this is fine. For corporate users with many tenants, this is an unbounded parallel burst against Microsoft Entra. It is not currently a known cause of failures, but it is the same anti-pattern that bit us in #685 with `estimateDocumentCount`.

## Proposal

Apply the shared `ConcurrencyLimiter` from `src/utils/concurrencyLimiter.ts` (added in #685). Suggested starting point: `concurrency: 5`, no inter-batch delay (this is a foreground UX path, latency matters). Both call sites should share a single limiter so the total parallelism is bounded across the wizard.

## Acceptance criteria

- Both call sites routed through a shared limiter.
- Behavior unchanged for users with a small tenant count.
- No regression in the account-selection / subscription-filtering wizards.

## References

- PR #685: introduces the limiter primitive.

Issue #668, opened 2026-05-21
Audit context menu commands for tree view multi-select support


tnaum-ms:
## Summary

The Connections View tree has `canSelectMany: true`, but most context menu commands only operate on a single item. When multiple items are selected, these commands either silently act on only the first item or fail. We need to either:

1. **Hide** single-item commands when multiple items are selected (add `&& !listMultiSelection` to `when`)
2. **Add multi-select support** where batch operations make sense

### How `listMultiSelection` works

VS Code provides the `listMultiSelection` context key, which is `true` when more than one item is selected in a tree view. Use it in the `when` clause of `view/item/context` menu entries in `package.json`:

```json
// Hide when multiple items are selected:
"when": "view == connectionsView && viewItem =~ /.../ && !listMultiSelection"

// Show ONLY when multiple items are selected:
"when": "view == connectionsView && viewItem =~ /.../ && listMultiSelection"
```

> **Note:** Only the `connectionsView` has `canSelectMany: true`. The `discoveryView` and Azure Resource views do not, so their commands are unaffected.

### Already handled

| Command | Status |
|---------|--------|
| Delete Connection (`removeConnection`) | Multi-select supported (#667) |
| Move to Folder (`moveItems`) | Multi-select supported |

#### Cluster-level (connectionsView only)

| Command | Recommendation |
|---------|----------------|
| Update Connection String (`updateConnectionString`) | Add `!listMultiSelection` |
| Update Credentials (`updateCredentials`) | Add `!listMultiSelection` |
| Rename Connection (`renameConnection`) | Add `!listMultiSelection` |

#### Cluster-level (shared across views)

| Command | Recommendation |
|---------|----------------|
| Create Database (`createDatabase`) | Add `!listMultiSelection` |
| Copy Connection String (`copyConnectionString`) | Add `!listMultiSelection` |
| Open Interactive Shell (`shell.open`) | Add `!listMultiSelection` |
| Data Migration (`accessDataMigrationServices`) | Add `!listMultiSelection` |

#### Database-level

| Command | Recommendation |
|---------|----------------|
| Create Collection (`createCollection`) | Add `!listMultiSelection` |
| Drop Database (`dropDatabase`) | Add `!listMultiSelection` — or add multi-delete support |
| Paste Collection (`pasteCollection` on database) | Add `!listMultiSelection` |
| Open Shell (`shell.open` on database) | Add `!listMultiSelection` |
[...]

Issue #659, opened 2026-05-21
Show index count on the Indexes folder node 🔢


tnaum-ms:
## Problem

The "Indexes" folder node in the tree view shows no indication of how many indexes a collection has. Users must expand the node to see its contents.

```
Current:
▼ my-collection
    ▶ Documents
    ▶ Indexes                        <-- no count

Expected:
▼ my-collection
    ▶ Documents
    ▶ Indexes                        (5 indexes)
```

## Expected Behavior

After the `IndexesItem` node is rendered, it should show the index count as its `description`. Since the index list is fetched when the node is expanded (in `getChildren()`), the count can be updated after the first expansion, or loaded lazily in the background.

## Development Hints

Since this is a `good first issue`, here are some hints on where to start:

**Pattern to follow:** `src/tree/documentdb/CollectionItem.ts`

`CollectionItem` implements a lazy loading pattern for document count that you can replicate. The approach uses:
- A cached count field with three states: `undefined` (not loaded), `number` (loaded), `null` (failed)
- A boolean guard to prevent duplicate fetches
- A fire-and-forget method that loads the count and calls `ext.state.notifyChildrenChanged(this.id)` to refresh the tree item

**Important consideration:** Unlike document counts, the index list is already fetched when the user expands the `IndexesItem` node (via `getChildren()` which calls `ClustersClient.listIndexes()`). You should investigate whether the count can be derived from that existing fetch rather than making an extra API call. Possible approaches:

1. **After-expansion update:** Cache the count after `getChildren()` runs and refresh the tree item to show it. This means the count appears after the first expansion.
2. **Lazy background loading:** Replicate the `CollectionItem` pattern with a separate background fetch. This shows the count before expansion but makes an extra API call.

Evaluate which approach makes more sense. Copying the lazy loading code from `CollectionItem` is fine for this issue. We can create a follow-up refactoring issue to deduplicate the pattern later. The focus here is the feature, not code reuse.

**File to modify:** `src/tree/documentdb/IndexesItem.ts`

In `getTreeItem()`, add:
```typescript
description: typeof this.indexCount === 'number'
    ? (this.indexCount === 1 ? l10n.t('1 index') : l10n.t('{0} indexes', this.indexCount))
    : undefined
```

The parent that creates `IndexesItem` is `CollectionItem.getChildren()` in `src/tree/documentdb/CollectionItem.ts`.

## Files to Modify

| File | Change |
|------|--------|
[...]

PR #702, opened 2026-05-29
perf(playground): replace fixed 2s sleep with TS server readiness polling

### LINKED ISSUE #638: Revisit: replace 2-second sleep before typescript.restartTsServer with a readiness signal
## Context

In [`ClustersExtension.ts`](https://github.com/microsoft/vscode-documentdb/blob/next/src/documentdb/ClustersExtension.ts), `ensureTsRestart` waits 2000 ms before issuing `typescript.restartTsServer`:

```ts
await new Promise((resolve) => setTimeout(resolve, 2000));
await vscode.commands.executeCommand('typescript.restartTsServer');
```

The comment explains: "Without this delay, the restart command can arrive while the server is still starting, causing a crash."

This is a magic number. On a slow or cold-start machine the TS server may still be initializing after 2 s, in which case the restart can fail. On fast machines the wait is unnecessary user-perceived latency.

Telemetry coverage for this code path was added in 0.8.1 (event `playground.tsPluginBootstrap`, property `stage = 'tsWait' | 'tsRestart'`), so we will get a real signal about failure rates once 0.8.1 is in users' hands.

## Suggested solutions

1. Replace the fixed 2 s delay with a polling loop that checks for a more reliable readiness signal. Candidates to investigate:
   - `vscode.extensions.getExtension('vscode.typescript-language-features')?.exports` once it becomes non null.
   - `vscode.languages.getLanguages()` returning `typescript` consistently.
   - Tracking `vscode.workspace.onDidOpenTextDocument` for the first `typescript` or `typescriptreact` document since extension activation.
2. Use exponential backoff with a hard cap (for example 200 ms, 500 ms, 1 s, 2 s, 4 s) and emit telemetry on each retry so we can correlate restart latency with reliability.
3. Investigate whether the official TS extension exposes an API surface to wait for `serverStarted` or to register plugins post hoc, so we can avoid the restart entirely.

## Definition of done

- The `tsWait` stage no longer relies on a fixed delay.
- `playground.tsPluginBootstrap` telemetry shows reduced failure rate at the `tsRestart` stage in the field.

hanhan761:
@
## Summary
Replace the fixed 2000ms `setTimeout` before `typescript.restartTsServer` with an exponential backoff polling loop.

### Before
```ts
await new Promise((resolve) => setTimeout(resolve, 2000));
```

### After
```ts
const pollDelays = [200, 500, 1000, 2000, 4000];
for (const delay of pollDelays) {
    await new Promise((resolve) => setTimeout(resolve, delay));
    const ext = vscode.extensions.getExtension("vscode.typescript-language-features");
    if (ext?.isActive && ext.exports) break; // ready
}
```

## Benefits
- **Fast machines**: returns after ~200ms instead of always waiting 2s
- **Slow/cold-start machines**: backoff up to ~7.7s (hard cap 10s) 
- **Telemetry**: `tsReadyAfterPoll` + `tsWaitDurationMs` for field correlation

## Issue
Fixes #638
@


tnaum-ms:
@hanhan761 Thank you for your contribution! Were you able to test this whether the pooling actually works? I remember having some issues during my initial run so that's it ended up as an issue for future work :)

tnaum-ms:
Thanks for the contribution, @hanhan761, and apologies for the delayed review.

We took a close look (including rebasing the branch locally onto current `main`) and unfortunately the change this PR advertises — replacing the fixed 2s `setTimeout` before `typescript.restartTsServer` with readiness polling / exponential backoff — **isn't present in the committed code**. The branch tip makes no changes to `src/documentdb/ClustersExtension.ts`; there is no `pollDelays` loop and no `tsReadyAfterPoll`/`tsWaitDurationMs` telemetry. `main` still has the original `setTimeout(2000)`, and the branch matches it exactly. It looks like the polling work was dropped during the force-push on 2026-05-29 while addressing the Copilot review.

What the branch still carries are unrelated bundled changes, most of which have since landed on `main` through other PRs:
- Index count on the Indexes node (#659) — already merged on `main`.
- Multi-select context-menu guards (#668) — already merged on `main`.
- Tenant sign-in throttling (#688) — duplicated by your dedicated PR #694.

Because the #638 fix is missing and the remaining content is superseded or duplicated, we're going to **close this PR**. Issue #638 stays open so the readiness-signal work can be picked up in a fresh, focused PR that actually edits `ClustersExtension.ts`. Thanks again!

PR #703, opened 2026-05-29
fix(query-insights): traverse shard branches in planner-tree helpers

### LINKED ISSUE #620: Query Insights: Planner-tree traversal helpers miss shard branches
## Summary

`findStageInPlan()` in `ExplainPlanAnalyzer.ts` and `collectPlannerStages()` in `StagePropertyExtractor.ts` only traverse `inputStage` and `inputStages` — they do **not** recurse through `shards` structures. In contrast, `traverseStages()` (execution-stage traversal for UI) already handles `shards`.

This means for sharded DocumentDB clusters, planner-stage signals like `isBitmap` can be missed when they appear under a shard branch. Only Stage 2 informational badges (bitmap, low-cardinality) are affected — Stage 3 (AI) receives the raw explain JSON and is unaffected.

## What needs to change

- Extend `findStageInPlan()` to recurse through `shards` structures (matching `traverseStages()` behavior).
- Extend `collectPlannerStages()` similarly.
- Add a sharded test case where `IXSCAN.isBitmap === true` appears under a shard branch.

## Related

- #522 — *Query Insights: Improve with Explain Schema* — should be tackled together, as formalizing the explain schema will inform how shard traversal should be structured.

## Origin

Identified during review of PR #616 (R-002 in `docs/analysis/pr-616-review.md`). Severity: Low.

hanhan761:
## Problem

`findStageInPlan()` and `collectPlannerStages()` only traversed `inputStage` and `inputStages` — they missed stages nested under `shards` structures in sharded cluster explain plans. This meant planner-stage signals like `isBitmap` (bitmap index detection) could be silently missed for sharded DocumentDB clusters.

In contrast, `traverseStages()` (execution-stage traversal for UI) already handled `shards`.

## Fix

- **`findStageInPlan()`** (`ExplainPlanAnalyzer.ts`): added `shards` array recursion, matching the pattern used by `traverseStage()` and `checkStageForSort()` elsewhere in the same file.
- **`collectPlannerStages()`** (`StagePropertyExtractor.ts`): added `shards` array recursion, matching `traverseStages()` in the same class.

## Tests

Two new test cases added:
1. **IXSCAN with isBitmap under shard path** — `SHARD_MERGE → shards → SINGLE_SHARD → FETCH → IXSCAN`
2. **Mixed planner/exec sharding** — planner is sharded but execution stats are flat

All 15 existing + 2 new tests pass. ✅

Closes #620

### DIFF
--- a/src/documentdb/queryInsights/ExplainPlanAnalyzer.addIndexStrategyAdvisories.test.ts
+++ b/src/documentdb/queryInsights/ExplainPlanAnalyzer.addIndexStrategyAdvisories.test.ts
@@ -349,7 +349,12 @@ describe('ExplainPlanAnalyzer.addIndexStrategyAdvisories', () => {
         });
 
         it('detects IXSCAN in planner shards even when exec stats use flat structure', () => {
-            // Edge case: planner is sharded, execution stats may not be
+            // Verifies that bitmap index detection works when the queryPlanner
+            // includes shard info (SHARD_MERGE > SINGLE_SHARD > IXSCAN) but
+            // executionStats use a simplified flat structure (FETCH > IXSCAN)
+            // without shard wrappers. This is important because production
+            // sharded clusters may return plan trees in either format, and the
+            // bitmap detector must be able to traverse both shapes correctly.
             const analysis = makeAnalysis();
             const explainResult: Document = {
                 queryPlanner: {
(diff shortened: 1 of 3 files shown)


tnaum-ms:
Thanks for this one. The shard branch traversal is a real gap and the direction here is right.

We are putting this on hold and scheduling it for the 0.10.1 release. The reason is testing. Doing this correctly means handling the different shapes of sharded explain output across platforms, and the current tests use simplified fixtures that do not match real sharded explain output, so we cannot validate it with confidence yet. We want a larger test setup before we commit to the change.

We expect to have that after our MongoDB Atlas discovery work ships, which will let us test against a wider range of platforms. We have marked the PR as on hold and moved it to the 0.10.1 milestone, and we will pick it back up then. Appreciate your patience.

PR #704, opened 2026-05-29
fix(tree): address Copilot review findings for index count + tenant sign-in + multi-select

hanhan761:
## Summary

Addresses all Copilot review findings on PRs #692 (index count), #693 (multi-select when clauses), and #694 (tenant sign-in limiter).

### Copilot review fixes (new commit)
- **Sort before caching**: combinedIndexes/indexes are sorted BEFORE assignment to this.cachedIndexes, so getChildren() never mutates the shared array in-place — eliminates the race window Copilot identified
- **Error logging**: listSearchIndexesForAtlas failures are now logged via ext.outputChannel.warn() in both fetchAndUpdateCount() and getChildren(), making transient network/auth errors diagnosable
- **Outer error logging**: fetchAndUpdateCount outer catch also logs failures via ext.outputChannel.warn()

### Previous commits (from ancestor branches)
- feat(tree): show index count on Indexes folder node (#692)
- fix(tree): hide single-item context menu commands when multiple items selected (#693)
- perf(azure): throttle parallel tenant sign-in checks with shared concurrency limiter (#694)
- fix(tree): address Copilot review findings for index count feature (initial fixes)

## Verification

- TypeScript compilation: zero errors in src/ directory
- All existing Copilot concerns addressed:
  - [x] In-place sort mutation -> sort-before-cache pattern
  - [x] Silent error swallowing -> ext.outputChannel.warn() logging
  - [x] Hardcoded singular/plural -> l10n.t() with localization bundle
  - [x] indexCount === null never retries -> guard uses typeof === number


tnaum-ms:
Thank you for this contribution! We apologize for the delayed review, we had the v0.9.0 release in the pipeline and it took up a lot of our capacity. We are closing this PR because the index count feature at its core was implemented on main via PR #714 as part of v0.8.1. The remaining concerns it covers (abort signal handling, logging, command visibility) are each worth addressing, but they will be easier to review in separate, focused PRs. We appreciate the work that went into capturing all those review findings!

Issue #641, opened 2026-05-13
Revisit: confirm parseShellBSON (Loose) backward compatibility with EJSON.parse for projection and sort


tnaum-ms:
## Context

In `ClustersClient.runFindQuery` / `streamFindQuery` and `ClusterSession.getCurrentFindQueryParamsForExecution`, projection and sort strings are now parsed with `@mongodb-js/shell-bson-parser` in `ParseMode.Loose`, replacing `EJSON.parse`.

The loose parser is a superset of strict EJSON and accepts mongosh style input such as unquoted keys and `ObjectId(...)`. Telemetry already counts parse failures via `QueryError('INVALID_PROJECTION' | 'INVALID_SORT')`, so post-release regressions are observable.

This issue tracks a focused compatibility pass to confirm that strict EJSON inputs that previously parsed still produce the same runtime values.

## Suggested verification

- Projection `{ "createdAt": { "$date": "2024-01-01T00:00:00Z" } }` produces an equivalent `Date`.
- Sort `{ "_id": -1 }` is unchanged.
- Nested extended JSON: `{ "$gt": { "$numberLong": "9007199254740992" } }`.
- Empty `{}` strings remain short-circuited.

## Definition of done

- A small fixture-driven unit test asserts equivalence between the old `EJSON.parse` output and the new `parseShellBSON` output for a representative set of strict EJSON inputs.

PR #708, opened 2026-05-29
Revisit: confirm parseShellBSON (Loose) backward compatibility with EJSON.parse for projection and sort

hanhan761:
## Summary\n- Add a shared parser that tries strict EJSON first, then falls back to loose shell BSON syntax.\n- Use the shared parser in ClustersClient and ClusterSession for projection/sort parsing.\n- Add regression tests for strict EJSON fixtures and loose shell BSON input.\n\n## Issue\nFixes #641\n\n## Verification\n- npx jest --selectProjects extension --runInBand src/documentdb/ClustersClient.test.ts\n- npx tsc -p . --noEmit

### DIFF
--- a/src/documentdb/ClusterSession.ts
+++ b/src/documentdb/ClusterSession.ts
@@ -4,7 +4,6 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { type FieldEntry } from '@documentdb-js/schema-analyzer';
-import { ParseMode, parse as parseShellBSON } from '@mongodb-js/shell-bson-parser';
 import * as l10n from '@vscode/l10n';
 import { EJSON } from 'bson';
 import { ObjectId, type Document, type Filter, type WithId } from 'mongodb';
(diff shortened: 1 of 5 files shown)


tnaum-ms:
Thanks for digging into this, and for the careful test fixtures.

After a deeper look we are going to close this one. The loose shell BSON parser is the same parser the driver uses at execution time, so it is our real baseline. Trying strict EJSON first and falling back introduces a subtle divergence: for extended JSON like {"$date": ...} or {"$oid": ...}, strict EJSON coerces to real Date and ObjectId types, while the loose parser (and therefore the driver) keeps them as plain objects. That means we could end up validating something different from what the query actually runs.

Since the loose parser already covers the syntax users type for projection and sort, the safest behavior is to keep using it directly. We really appreciate the investigation here, it helped us confirm the right contract and we would rather not add a second parsing path that can drift from the driver.

PR #712, opened 2026-05-30
fix: detectMethodArgContext skips parens inside strings and regex literals

### LINKED ISSUE #710: Shell completions should ignore parentheses inside strings and regex literals
### Problem

`ShellCompletionProvider.detectMethodArgContext()` determines whether the cursor is inside a collection method call by scanning backward and counting raw `(` / `)` characters. The scanner does not ignore parentheses that appear inside quoted strings or regex literals.

That can make the provider miss the real method-call boundary and suppress completions while the user is still editing a valid query object.

### Repro examples

```js
db.users.find({ note: ")", na| })
db.users.find({ name: /foo(bar)/, ag| })
```

At the cursor, field-name completions should still be available for the `find()` filter object.

### Expected behavior

Parentheses inside string literals, regex literals, and comments should not affect method-argument context detection.

### Current behavior

A literal `)` inside input such as `")"` or `/foo(bar)/` can change the depth calculation and cause completions to disappear.

### Notes

This is called out in `docs/ai-and-plans/shell-autocompletion/9-shell-autocompletion-review.md` under review item `#21`.

hanhan761:
## Summary

`ShellCompletionProvider.detectMethodArgContext()` now skips parentheses inside string literals and regex literals when scanning backward for the unmatched `(`. Similarly, `playgroundContextDetector.detectMethodArgContext()` masks all non-code regions (strings, line comments, block comments) before the paren scan. Previously, punctuation inside strings, regexes, or comments could shift the open/close paren depth count and cause completions to disappear.

### Changes in `ShellCompletionProvider.ts`
- Added string-aware backward scan: quotes trigger `skipStringBackward()` to jump past the entire string
- Added regex-aware backward scan: `/` in regex-start context triggers `skipRegexBackward()` to jump past the entire regex
- Added `skipStringBackward()` helper (tracks escape sequences)
- Added `skipRegexBackward()` helper (tracks escape sequences)
- Both helpers return the position before the opening delimiter

### Changes in `playgroundContextDetector.ts`
- Added `maskNonCodeRegions()` helper that blanks the contents of string literals, line comments (`//`), and block comments (`/* */`) while preserving string length and newline positions
- `detectMethodArgContext()` now passes text through `maskNonCodeRegions()` before the paren scan so only real-code parentheses are counted

### Test coverage
- `db.users.find({ note: ) , na` -> correctly detects `method-argument`
- `db.users.find({ name: /foo(bar)/, ag` -> correctly detects `method-argument`
- `db.users.find({ active: true /* ) */, na` -> correctly detects `method-argument` (comment parens ignored)
- `db.users.find({ active: true // )
na` -> correctly detects `method-argument` (line comment parens ignored)

## Issue

Fixes #710

## Verification
- All 553 shell tests pass (17 suites)
- All 89 completion provider tests pass
- Two new TDD tests for comment-inside-call scenarios pass

### DIFF
--- a/src/documentdb/query-language/playground-completions/playgroundContextDetector.ts
+++ b/src/documentdb/query-language/playground-completions/playgroundContextDetector.ts
@@ -272,25 +272,29 @@ function scanMemberChain(text: string, offset: number): MemberChain | null {
 /**
  * Detects if the cursor is inside a method argument: `db.users.find({ | })`
  * This is a simpler check — looks for an unmatched `(` before the cursor.
+ *
+ * Before counting parentheses we mask out strings and comments (see
+ * `maskNonCodeRegions`) so that punctuation such as `)` inside a string,
+ * a `//` line comment, or a `/* *\/` block comment cannot be mistaken for
+ * real code and throw off the paren-depth count.
  */
 export function detectMethodArgContext(text: string, offset: number): MethodCallInfo | null {
+    // Mask strings and comments so their punctuation does not affect the scan.
+    // Indexes are preserved, so positions found here map back onto `text`.
+    const scan = maskNonCodeRegions(text);
+
     let pos = offset - 1;
     let depth = 0;
 
     // Scan backward looking for unmatched `(`
     while (pos >= 0) {
-        const ch = text[pos];
-
-        // Skip string literals
-        if (ch === '"' || ch === "'") {
-            pos = skipStringBackward(text, pos);
-            continue;
-        }
+        const ch = scan[pos];
 
         if (ch === ')') depth++;
         else if (ch === '(') {
             if (depth === 0) {
-                // Found the unmatched `(` — now read the method name before it
+                // Found the unmatched `(` — now read the method name before it.
+                // Names live in real code, so we read them from the original text.
                 let methodPos = pos - 1;
                 while (methodPos >= 0 && isWhitespace(text[methodPos])) {
                     methodPos--;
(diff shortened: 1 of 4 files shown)


tnaum-ms:
Thank you again for this contribution. We are genuinely amazed by both the quality and the volume of improvements you have submitted. We also apologize for the delayed review while the 0.9.0 release was in flight. This fix improves completion correctness in a careful and practical way.

tnaum-ms:
Small follow-up after rebasing onto main.

What was spotted: the backward scan in detectMethodArgContext counted every parenthesis it saw, including parens that live inside comments. So an input like db.users.find({ active: true /* ) */, na would see the ) in the comment, treat it as real code, and throw off the open/close paren count. The result was that method argument completion stopped working when a comment with parens appeared inside the call. The same applied to line comments such as // ).

What was fixed: before counting parentheses, the text is now passed through a small helper (maskNonCodeRegions) that blanks out the contents of string literals, line comments, and block comments while keeping the same length and newlines. The paren scan then runs on that masked copy, so only real code parens are counted. Method and collection names are still read from the original text, so nothing else changes.

Why: comments and strings should never affect how we match parentheses. The earlier version already handled strings, but comments were still being read as code. Masking both in one place keeps the logic simple and covers all three cases (strings, line comments, block comments) consistently.

Two new TDD tests cover the comment cases and now pass, along with the full suite.

tnaum-ms (approved):
Thank you again for this contribution, and for the quality and volume of the work you have been sending our way. We are genuinely amazed by it. Apologies for the delay in reviewing while we were focused on shipping 0.9.0.

Approved after rebasing onto main and a full validation pass (lint, the complete test suite, and a clean build). I pushed one small follow-up so that parentheses inside comments are ignored during method argument detection. Details are in the comment above.

sajeetharan (approved):

sajeetharan (approved):

PR #713, opened 2026-05-30
fix(autocomplete): referenceText must not emit invalid MQL for unsafe field names

### LINKED ISSUE #709: referenceText should not emit invalid MQL for unsafe field names
### Problem

`toFieldCompletionItems()` currently emits `referenceText` as `$` + field path for every field. That breaks for field names that are not safe as a plain `$field` reference, such as:

- `order-items`
- `my field`
- `say"hi"`

Expected: unsafe names should not produce a broken aggregation reference.
Current: the completion data exposes an invalid `referenceText`, which can lead to bad MQL if consumed directly.

### Notes

This was documented as future work in `src/utils/json/data-api/autocomplete/future-work.md`.

Jacquelinezhong:
## Summary

Closes #709

`toFieldCompletionItems()` previously emitted `referenceText` as `` `$${entry.path}` `` for every field, which produces invalid MQL for field names containing special characters (e.g., `$order-items`, `$my field`).

**Changes:**
- Check each dot-separated path segment against `JS_IDENTIFIER_PATTERN` individually (not the full path, so nested paths like `address.city` keep their valid `$address.city` form)
- Flat field with an unsafe segment (e.g., `order-items`) → `{ $getField: "order-items" }`
- Nested path with an unsafe segment (e.g., `order.item-count`) → `referenceText` omitted; chained `$getField` form is complex and tracked as a follow-up in `future-work.md`
- `FieldCompletionData.referenceText` widened from `string` to `string?` so the tRPC client-side inferred type stays consistent
- `future-work.md` item 2 marked as resolved

## Behaviour table

| Field path | Before | After |
|---|---|---|
| `age` | `$age` ✅ | `$age` ✅ |
| `address.city` | `$address.city` ✅ | `$address.city` ✅ |
| `order-items` | `$order-items` ❌ | `{ $getField: "order-items" }` ✅ |
| `my field` | `$my field` ❌ | `{ $getField: "my field" }` ✅ |
| `say"hi"` | `$say"hi"` ❌ | `{ $getField: "say\"hi\"" }` ✅ |
| `order.item-count` | `$order.item-count` ❌ | `undefined` ✅ (no invalid MQL emitted) |

## Test plan

- [x] `toFieldCompletionItems.test.ts` — 13 tests pass (3 new tests added for unsafe flat fields, unsafe nested paths, and safe nested paths)
- [x] `tsc --noEmit` — no type errors

### DIFF
--- a/src/utils/json/data-api/autocomplete/future-work.md
+++ b/src/utils/json/data-api/autocomplete/future-work.md
@@ -15,41 +15,16 @@ Tests cover dashes, brackets, digits, embedded quotes, and backslashes.
 
 ---
 
-## 2. `referenceText` is invalid MQL for special field names
+## ~~2. `referenceText` is invalid MQL for special field names~~ ✅ RESOLVED
 
-**Severity:** Medium — will generate broken aggregation expressions
-**File:** `toFieldCompletionItems.ts` — `referenceText` construction
-**When to fix:** Before the aggregation completion provider is wired up
+**Resolved in:** PR #709 fix
 
-### Problem
+Per-segment check (`JS_IDENTIFIER_PATTERN` applied to each dot-separated segment):
+- All segments valid → `$field.path` (e.g., `$address.city`)
+- Flat field with unsafe name → `{ $getField: "name" }` (e.g., `{ $getField: "order-items" }`)
+- Nested path with unsafe segment → `undefined` (chained `$getField` deferred; see item 3)
 
-`referenceText` is always `$${entry.path}` (e.g., `$address.city`). In MQL, the `$field.path` syntax only works when every segment is a valid identifier without dots, spaces, or `$`. For field names like `order-items`, `a.b`, or `my field`, the `$` prefix syntax produces invalid references.
-
-### Examples
-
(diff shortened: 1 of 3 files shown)


tnaum-ms:
Thank you for this contribution, and apologies for the delay in responding.

With the larger volume of PRs coming in recently, we missed this submission and started reviewing another PR that addresses the same issue (#717). That was an oversight on our side, not a reflection of the quality of your work here. Your PR was actually the earlier submission, it was clean and isolated, and importantly it avoided emitting invalid MQL by returning an undefined reference for the nested unsafe case rather than producing a wrong expression.

While finalizing #717 we added a few improvements on top of the shared approach:

- nested paths with an unsafe segment now produce a correct chained `{ $getField: { field, input } }` reference instead of being omitted
- field names containing `$` are routed through `$getField` (a plain `$`-prefix would be misread as a variable)
- expanded test coverage and a scope cleanup of that branch

Because #717 builds out those additional cases, we are going to merge #717 to close out #709. This is not a quality judgment against your PR, the two overlapped and we are consolidating on one. We appreciate the time and care you put into this, and we hope to see more contributions from you.

Closing this in favor of #717, with thanks.

Issue #244, opened 2025-09-05
Introduce "Service Discovery" for MongoDB (RU) resources


tnaum-ms:
#### 1.1 Create Main Discovery Provider
**File**: `src/plugins/service-azure-mongo-ru/AzureMongoRUDiscoveryProvider.ts`

**Requirements**:
- Extend `Disposable` and implement `DiscoveryProvider` interface
- Copy structure from AzureDiscoveryProvider.ts
- Modify properties:
  - `id = 'azure-mongo-ru-discovery'`
  - `label = l10n.t('Azure Cosmos DB for MongoDB (RU)')`
  - `description = l10n.t('Azure Service Discovery for MongoDB RU')`
  - `iconPath = new ThemeIcon('azure')`
- Update `getLearnMoreUrl()` to return RU-specific URL (e.g., `'https://aka.ms/vscode-documentdb-discovery-providers-azure-ru'`)
- **Reuse existing infrastructure**: Import and use the same `AzureSubscriptionProviderWithFilters` from `'../api-shared/azure/AzureSubscriptionProviderWithFilters'`
- **Reuse existing filtering**: Import and use `configureAzureSubscriptionFilter` from `'../api-shared/azure/subscriptionFiltering'`
- Reference new wizard steps and tree items (to be created)

#### 1.2 Create Context Properties Enum
**File**: Same file as 1.1

**Requirements**:
- **Reuse existing enum**: Import `AzureContextProperties` from `'../service-azure/AzureDiscoveryProvider'`
- Do NOT create a new enum - use the existing one for consistency and shared subscription filtering

#### 1.3 Create Discovery Tree Structure

**File**: `src/plugins/service-azure-mongo-ru/discovery-tree/AzureMongoRUServiceRootItem.ts`
- Copy from `service-azure/discovery-tree/AzureServiceRootItem.ts`
- Update contextValue to `'azureMongoRUService'`
- Update child creation to use RU-specific subscription items
- Update labels/descriptions to reference "MongoDB (RU)" instead of "MongoDB (vCore)"

**File**: `src/plugins/service-azure-mongo-ru/discovery-tree/AzureMongoRUSubscriptionItem.ts`
- Copy from `service-azure/discovery-tree/AzureSubscriptionItem.ts`
- Update contextValue to `'azureMongoRUSubscription'`
- Update child creation to use RU-specific resource items
- Modify `getChildren()` to use CosmosDB Management Client instead of ResourceManagementClient

[...]

Issue #604, opened 2026-04-23
SchemaStore: Add memory ceiling with LRU eviction for _analyzers Map


tnaum-ms:
## Problem

The `SchemaStore._analyzers` Map grows unbounded. If a user queries many collections across many clusters in a long session, memory usage increases without limit.

## Current Behavior

- `_analyzers` is a `Map<string, SchemaAnalyzer>()` keyed by `{clusterId}::{db}::{coll}`
- New analyzers are created on each `addDocuments()` call for new collections
- No eviction strategy exists

## Proposed Solution

Add a soft limit (e.g., 100 entries) with time-based eviction:
- Evict entries older than 1 hour, checked every hour
- Consider LRU eviction if the limit is reached before the time threshold

## Telemetry

Telemetry has been added (commit in PR #573) to measure:
- `maxCollectionCount` — peak analyzer count
- `currentCollectionCount` — current snapshot
- `distinctClusterCount` — number of distinct clusters
- `currentTotalDocuments` / `currentTotalFields` — total docs and fields

Telemetry reports periodically (every call for the first 10 `addDocuments` calls, then every 5th). This data will help assess actual usage patterns before implementing eviction.

## Priority

Optional — measure first via telemetry, then decide on eviction parameters.

## References

- PR #573 review issue M2
- `src/documentdb/SchemaStore.ts`

PR #725, opened 2026-06-02
chore: remove applicationinsights-web-basic override after upstream fix release (fixes #614)

### LINKED ISSUE #614: chore: remove applicationinsights-web-basic override after upstream fix is released
## Summary

We pin \`@microsoft/applicationinsights-web-basic\` to \`~3.3.4\` via an npm override in \`package.json\` to work around an upstream bug in \`@vscode/extension-telemetry\`.

## Root Cause

\`@vscode/extension-telemetry\` (all versions through 1.5.1) imports \`@microsoft/applicationinsights-common\` at runtime but does not declare it as a direct dependency. It relied on it arriving transitively through \`@microsoft/applicationinsights-web-basic\`. In version 3.4.x, \`applicationinsights-web-basic\` removed \`applicationinsights-common\` from its dependency tree, breaking any consumer that allows \`^3.3.x\` to resolve to \`3.4.x\`.

## Upstream Fix

- **Issue**: https://github.com/microsoft/vscode-extension-telemetry/issues/244
- **Fix PR**: https://github.com/microsoft/vscode-extension-telemetry/pull/245 (merged, bumps to v1.5.2)
- **Status**: v1.5.2 is **not yet published to npm** as of 2026-04-25

## ⚠️ Side Effect: VSIX Size Increase

The override has an important size side effect introduced in #613. When \`package-lock.json\` is regenerated while the override is active, \`@microsoft/1ds-core-js\` resolves to \`4.4.1\`, which requires \`applicationinsights-core-js@3.4.1\`. This conflicts with the pinned \`web-basic@3.3.11\`, which requires \`applicationinsights-core-js@3.3.x\`. npm resolves the conflict by creating nested copies of the ApplicationInsights packages under \`@vscode/extension-telemetry/node_modules/\` and \`@microsoft/applicationinsights-common/node_modules/\`. Since webpack bundles these packages (they are not in \`externals\`), both the \`3.4.1\` and \`3.3.11\` variants end up in the extension bundle, adding ~100–120 KB (compressed) to the VSIX.

Removing the override will collapse these back to a single shared resolution and recover the size.

## Action Items

- [ ] Monitor the release of \`@vscode/extension-telemetry@1.5.2\` on npm
- [ ] Once released, verify that \`@microsoft/vscode-azext-utils\` picks up the new version
- [ ] Remove the \`@microsoft/applicationinsights-web-basic\` override from \`package.json\`
- [ ] Regenerate \`package-lock.json\` and confirm the nested copies are gone
- [ ] Run tests to confirm everything works without the pin
- [ ] Verify VSIX size returns to baseline

## Notes

- \`vscode-cosmosdb\` has the same override applied: \`"@microsoft/applicationinsights-web-basic": "~3.3.11"\`
- Even upgrading to \`@microsoft/vscode-azext-utils@4.x\` does not help today since it depends on \`@vscode/extension-telemetry@^1.5.0\` which resolves to 1.5.1 (still broken)
- Size analysis for the PR that introduced the override: #613

hanhan761:
## Summary

The `@microsoft/applicationinsights-web-basic` override was pinned to `~3.3.4` as a workaround for `@vscode/extension-telemetry` issue [#244](https://github.com/microsoft/vscode-extension-telemetry/issues/244). The upstream fix was released in v1.5.2, so the override is no longer needed.

## Changes

- Removed `@microsoft/applicationinsights-web-basic: ~3.3.4` from `overrides` in `package.json`
- Updated the `//overrides` comment to remove the applicationinsights mention

## Issue

Fixes #614

## Verification

- `npm install` completed successfully — dependencies resolve correctly without the override
- Only `package.json` changed (2 insertions, 3 deletions)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5157,47 +5157,31 @@
       }
     },
     "node_modules/@microsoft/applicationinsights-channel-js": {
-      "version": "3.3.11",
-      "resolved": "https://registry.npmjs.org/@microsoft/applicationinsights-channel-js/-/applicationinsights-channel-js-3.3.11.tgz",
-      "integrity": "sha512-0ex/mxTf5R+P5WSvdU8Hgbeg8VzQ0XvcnjKQdmQy05ycScnKevt8an3DEPikOFqOKDi59L5hUETZlcdhesnVtg==",
+      "version": "3.4.2",
+      "resolved": "https://registry.npmjs.org/@microsoft/applicationinsights-channel-js/-/applicationinsights-channel-js-3.4.2.tgz",
+      "integrity": "sha512-Q7Q9gV45WSgSmOP2abu0y9tdbFh8sPELMgKUCDy62FsNd3xmE8X3zLtkzc7mcXPlpTeoDXwMCYjCRIAYd6d2lQ==",
       "license": "MIT",
       "dependencies": {
-        "@microsoft/applicationinsights-common": "3.3.11",
-        "@microsoft/applicationinsights-core-js": "3.3.11",
+        "@microsoft/applicationinsights-core-js": "3.4.2",
         "@microsoft/applicationinsights-shims": "3.0.1",
         "@microsoft/dynamicproto-js": "^2.0.3",
-        "@nevware21/ts-async": ">= 0.5.4 < 2.x",
-        "@nevware21/ts-utils": ">= 0.11.8 < 2.x"
-      },
-      "peerDependencies": {
-        "tslib": ">= 1.0.0"
-      }
-    },
-    "node_modules/@microsoft/applicationinsights-channel-js/node_modules/@microsoft/applicationinsights-common": {
-      "version": "3.3.11",
-      "resolved": "https://registry.npmjs.org/@microsoft/applicationinsights-common/-/applicationinsights-common-3.3.11.tgz",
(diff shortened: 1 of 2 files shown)


tnaum-ms:
Thanks for picking this up, @hanhan761! The core change is exactly right and the prerequisites are all met -- extension-telemetry 1.5.2 is out and vscode-azext-utils 4.1.x has already updated its dependency range to pick it up.

We did a small cleanup before merging. The branch had two independent changes bundled together: the override removal (issue #614) and the SchemaStore LRU eviction (issue #604). The LRU work had already landed on main separately, so keeping it here would have created a conflict. We also noticed the package-lock.json was not updated in the PR -- without regenerating it, the nested duplicate AI packages would have stayed in the lockfile and the bundle size improvement would not have been visible.

We force-pushed a single clean commit that covers only issue #614: removes the override from package.json, updates the comment, and includes the regenerated package-lock.json. The lockfile now resolves applicationinsights-web-basic to a single ^3.4.1 entry (was two nested copies), which recovers the expected ~100-120 KB from the VSIX.

One note for future PRs: we noticed the same pattern of bundling unrelated changes in your other open PRs as well -- sorry for not flagging it sooner. Keeping each PR focused on one issue makes reviews much easier, both for confirming the change matches the issue requirements and for isolating any regressions. No worries at all, just something worth keeping in mind going forward.

Everything passes and we will merge this shortly. Great contribution!

sajeetharan (approved):

tnaum-ms (approved):

PR #790, opened 2026-07-09
documentdb local ux review

tnaum-ms:
ux


tnaum-ms:
Closing as superseded by #876.

This PR was a review artifact rather than a change: a 1144-line UX review plus one wording tweak (item 30), which has since shipped independently. It has been conflicting against `release/0.10.0` for a while, so rather than rebase a document whose status column is now wrong in eight places, here is where its findings actually went.

**Shipped** (verified on the #876 branch):

| # | Item | Where |
| --- | --- | --- |
| 2 | External container delete → silent no-op on Start | `ensureActionable()` distinguishes missing / foreign / wrong-state; regression test added |
| 4 | Tree nodes misused to display errors | `createErrorRecoveryChildren()` + `buildPreflightChildren()` render actions, not error text |
| 6 | "Delete deletes data" vs "recreate" clash | Missing tooltip now states both outcomes explicitly |
| 7 | Copy Connection String silently copies the password | Routed through the shared with/without-password QuickPick |
| 8 | Delete bypasses the shared confirmation | Now uses `getConfirmationAsInSettings` |
| 9 | Only ever delete containers we created | Label-based `isOwnedContainer()` plus a `refused` outcome |
| 23 | A user delete must drop the data volume | `deleteContainer()` removes the volume and warns if it cannot |
| 30 | Empty-state row reword | Shipped independently |

**Resolved differently from the recommendation:**

- **Item 1** (credential-missing dead end). The review recommended exposing **Delete Container** on that row. The shipped behaviour deliberately does not: the row routes to the Quick Start webview, where the destructive choice is explained in context, and a test pins the `state_needsAttention` token. Recording this as a decision rather than leaving it open.

**Carried forward:**

- **Item 20** (running row lacks regular cluster commands) — being addressed in #876 by opting the safe commands in, rather than granting the row the full cluster context value. Storage-targeting commands (rename, move, remove, update credentials/connection string, data migration) stay off, because the managed instance is not a stored connection.
- **Item 26** (tree state transitions are not announced) — now #878, milestone 0.10.1.
- **Items 10 / 11** (legacy migration folder, "local connection" concept) — covered by #870.
- **Recovery affordances** (open idea O4) — folded into #877.

Still open and not tracked anywhere yet: item 25 (should the container stop when VS Code closes), items 22 / 24 (ephemeral option, volume-to-image linkage), and the webview-side set (3, 5, 12–19, 21, 27–29). Those can be raised individually if and when they matter.

Issue #741, opened 2026-06-16
Kubernetes discovery: cap API calls at 30s and show retry error nodes (no built-in client timeout)


tnaum-ms:
## Summary

Kubernetes service discovery currently issues `@kubernetes/client-node` API calls **without any request timeout**. If the Kubernetes API server hangs the socket (VPN down, packet blackhole, wrong server URL behind a slow proxy), the discovery tree spins **indefinitely** (or until the OS TCP timeout, which can be minutes) with no way for the user to tell discovery from a hang.

This is the follow-up for finding **2.2** in the PR #621 pre-merge review (`docs/ai-and-plans/PRs/621-kubernetes-discovery/pre-merge-code-review.md`).

## Decision (agreed)

- **Cap every Kubernetes discovery API call at 30 seconds.**
- On timeout, surface the existing **error node with the "Click here to retry" action** (same path already used for RBAC/network failures) — i.e. fail into the established retry/error-node UX, not a silent hang.
- **Do NOT** wire up `AbortController`/cancellation on node collapse. Collapsing the tree node will not abort in-flight requests; the only behavior we add is the 30s cap + retry error node.

## Background: why this is needed

A research pass against the **actual shipped `@kubernetes/client-node` v1.4.0 code** in `node_modules` confirmed:

- **No built-in default request timeout** on the generated API clients (`CoreV1Api`, `CustomObjectsApi`, …). `createConfiguration()` has no timeout field; `makeApiClient()` passes none; the auth/agent hook sets none.
- The transport is **`node-fetch` v2** (not undici), called with only `method/body/headers/signal/agent` — **no `timeout` passed**. So even undici's `headersTimeout`/`bodyTimeout` (~300s) defaults **do not apply**.
- **Connect blackhole** (SYN dropped, no RST) → hangs until the **OS TCP connect timeout** (~20s to >2min). **Connected-but-silent server** → hangs **indefinitely**.
- The generated API methods **do not accept an `AbortSignal` argument**. The `timeoutSeconds?` param on `listNamespace(...)` is the Kubernetes **server-side** list/watch query parameter — it does nothing on a connect blackhole.
- The supported way to inject a signal is a **client-wide `pre` middleware** on the API client built from `KubeConfig` that calls `ctx.setSignal(AbortSignal.timeout(ms))`. `AbortSignal.timeout(ms)` covers both the connect phase and the awaiting-response phase, so a single 30s signal bounds the whole operation.

## Suggested implementation

[...]

PR #802, opened 2026-07-21
fix(kubernetes): time out API requests

guanzhousongmicrosoft:
## Summary

- Cap generated Kubernetes API requests at 30 seconds.
- Route timeout failures through the existing retry-node experience instead of leaving the tree spinning.
- Allow namespace pre-scan workers to continue after an individual namespace times out.

## Implementation

- Add a small timeout helper with a fresh `AbortSignal.timeout(30_000)` for every request.
- Construct `CoreV1Api` and `CustomObjectsApi` clients with the same kubeconfig authentication and base server as `KubeConfig.makeApiClient`, plus timeout middleware.
- Normalize aborted requests as `KubernetesApiTimeoutError` for clear logging, telemetry, and user-facing timeout messages.
- Preserve best-effort credential Secret lookup and the existing port-forward lifecycle.

## Review follow-up

- Raw `AbortError` / `TimeoutError` values are converted to the timeout type only at known generated-client API boundaries. Unrelated port-forward aborts retain their original cause.
- Already-wrapped timeout errors preserve their full operation and namespace context.
- Tree timeout UX checks the typed error before message-text fallbacks, so localization cannot change classification.
- Added focused timeout-helper and port-forward regression tests.

## User experience

- Context and namespace timeouts surface the existing **Click here to retry** action.
- A timed-out namespace remains expandable for retry while successfully scanned namespaces continue to render.
- The timeout is fixed at 30 seconds; this does not add cancellation on tree collapse or a new setting.

## Validation

- `npm run l10n`
- `npm run prettier-fix`
- `npm run lint`
- `npx jest --no-coverage`
- `npm run build`

Addresses #741.

### DIFF
--- a/src/plugins/service-kubernetes/discovery-tree/KubernetesContextItem.test.ts
+++ b/src/plugins/service-kubernetes/discovery-tree/KubernetesContextItem.test.ts
@@ -470,7 +470,7 @@ describe('KubernetesContextItem', () => {
 
         it('should identify a namespace-list timeout and retain the retry action', async () => {
             (vscode.window.showErrorMessage as jest.Mock).mockClear();
-            const timeoutError = new Error('Operation timed out after 30 seconds.');
+            const timeoutError = new Error('localized deadline message');
             timeoutError.name = 'KubernetesApiTimeoutError';
             mockListNamespaces.mockRejectedValue(timeoutError);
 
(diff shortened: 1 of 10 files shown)


guanzhousongmicrosoft:
Validated all five review findings against the current implementation and runtime:

1. **AbortError over-classification — valid, fixed.** Raw aborts are now branded only inside known timed generated-client API boundaries. A port-forward setup `AbortError` keeps its actual cancellation message.
2. **Localized substring classification — valid, fixed.** Context and namespace tree UX now checks `KubernetesApiTimeoutError` before retaining text matching as a fallback for ordinary network errors.
3. **Context stripping — valid, fixed.** Already-wrapped timeout errors keep their full contextual message through nested handlers and diagnostics.
4. **Duplicated `isRecord` — not changed.** There is no shared repository helper; importing the copy from `kubernetesClient.ts` would reverse the dependency and create a cycle. Extracting a new utility for this four-line local guard would add scope without meaningful value.
5. **Uncleared `AbortSignal.timeout` timer — not changed.** Node 22.18 implements this with an unref'd timer, a `WeakRef`, and `FinalizationRegistry` cleanup when the signal becomes unreachable. It neither keeps the process alive nor necessarily lingers for 30 seconds after a completed request; replacing it would add bespoke lifecycle complexity without a reliability benefit.

Added regression coverage for the accepted findings and reran localization, formatting, lint, the full Jest suite, and the TypeScript build.

PR #849, opened 2026-08-04
Redesign Local Quick Start as a wizard and fix the alternating Retry

tnaum-ms:
Redesigns the Local Quick Start webview as a four-step wizard and fixes a provisioning race that made `Retry setup` work only on every second click.

Targets `feature/local-quickstart`, not `main`: this branch is 5 commits ahead of that branch and 0 behind. Against `main` the diff would be 451 commits.

## The wizard

Four steps — Introduction, Configure, Set up, Done — sharing the chrome the Atlas credentials view already established. The responsive breadcrumb was extracted into `src/webviews/components/wizard/WizardBreadcrumb.tsx` and both views now use it, so a step indicator is no longer implemented twice.

**Introduction** states what will happen before anything is downloaded or created, as a numbered plan mirroring the setup stages.

**Configure** is a Fluent `Table` with no header row: a fixed label column, a value column that takes the leftover space, and a narrow action column of icon-only buttons labelled by their tooltips. Editors for the port, image tag, and credentials expand under the row, indented to the value column the way the index list indents its detail row.

**Set up** shows the five provisioning stages, each carrying evidence of what was actually observed rather than a transient log. A collapsed panel reports everything the Docker readiness check established — eleven facts, each with a note on how it was established, including why a fact is still unknown when the daemon never answered.

## Docker failures have exactly one home

A Docker problem is reported beside the stage that hit it, never on a screen of its own and never in two places at once. All the recovery affordances are reachable from there: `Check Docker again`, `Start Docker`, `Copy` for the recovery command, `Continue anyway`, `Wait longer`, and `Start over`.

`Check Docker again` re-runs only the Docker probe. It never starts or resumes provisioning: passing clears the blocker in place, and the footer then offers `Continue setup`.

## The alternating Retry — root cause and fix

`QuickStartService.provision` reports every terminal failure by buffering it into `terminalEvent`, letting `finally` clear the `provisioning` guard, and yielding it afterwards. The Docker-readiness failure was the one path that yielded in place, from inside the `try`. The generator then sat suspended at that `yield` with `provisioning` still set, while the webview already showed the failure and re-enabled the button. The next click unsubscribed the old stream and immediately subscribed a new one, which reached the guard before the old generator had unwound and came back with `Setup is already in progress.`. The click after that worked, because by then the unwind had finished.

Fixed by routing the readiness failure through the same buffered path — a typed `DockerNotReadyError` caught by the existing `catch`. The emitted event is unchanged; only its timing moved. Covered by a regression test that starts a second `provision()` at the instant the terminal event arrives.

## Motion

`Collapse` from `@fluentui/react-motion-components-preview`, the same component the Create Index drawer uses. Editors and the Docker blocks expand into place when they arrive and keep their content while collapsing away; anything already present at mount renders instantly, so opening a step never shifts. The Docker check row reserves its evidence line rather than animating it.

## Also

- The tree's empty-state row reads `Click here to set up DocumentDB Local`; the `Learn more…` link is gone.
- Two deferred items are recorded under **Known follow-ups** in the design decisions document: the unsubscribe/resubscribe handshake in `runStream`, which is correct today only because the service buffers its terminal events, and whether the error row belongs in the tree at all.
- `docs/ai-and-plans/live-preview-playwright-future-work.md` writes up the technique used throughout this work — rendering the production webview bundle in a plain browser and driving it with Playwright. It caught a misaligned icon, a code block punching through an error tint, a stretched input, and a Switch whose thumb had been pushed out of its track by an inherited `text-align`.

## Verification

`npm run l10n`, `npm run prettier-fix`, `npm run lint`, `npx jest --no-coverage` (199 suites / 3292 tests), and `npm run build` all pass. Layout, focus order, and accessible names were checked in the browser down to a 304 px viewport.


tnaum-ms:
@guanzhousongmicrosoft FYI: this is where the redesign was done, we'll discuss on the next sync.

Issue #858, opened 2026-08-04
Manual localhost connection creation does not detect existing DocumentDB Local entries


tnaum-ms:
Found during the `0.10.0-bug-bash-1` bug bash.

## Summary

When a user manually adds a localhost connection, the extension does not check whether it already has a matching DocumentDB Local connection. This allows duplicate entries for the same local service.

## Steps to reproduce

1. Start DocumentDB Local through Local Quick Start so its connection is present in the Connections view.
2. Manually add a connection that targets the same localhost endpoint.
3. Complete the connection flow.

## Expected

The extension detects the matching DocumentDB Local entry and prevents the duplicate, focuses the existing connection, or clearly asks the user whether to retain a separate configuration.

## Actual

A second connection can be added for the same local endpoint.

## Impact

Duplicate local connections make the Connections view harder to understand and can lead users to manage the wrong entry.

## Suggested fix

Before saving a manually added localhost connection, compare its normalized endpoint with existing DocumentDB Local entries. Account for equivalent host forms such as `localhost`, `127.0.0.1`, and IPv6 loopback where applicable, along with the port and relevant connection options.

tnaum-ms:
Additional bug-bash context: checked all issues created on 2026-08-04 and searched existing localhost connection reports. No matching duplicate-detection issue was found. The intended comparison should normalize loopback host aliases and retain enough connection settings to avoid collapsing genuinely different configurations.

tnaum-ms:
A user might really want to have another connection saved for the 'localhost' connection, so we should allow it. We should have a modal dialog that explains what's happening, and offer actions:
- open the one in DocumentDB Local,
- save anyway.

tnaum-ms:
Fixed by #866, merged into `feature/local-quickstart` as 2803a385: adding a localhost connection by hand now detects the managed instance and offers Open Existing / Add Anyway / cancel, with normalized loopback host comparison.

Closing by hand rather than automatically: #866 targeted `feature/local-quickstart`, not the default branch, so GitHub ignores closing keywords and creates no link. Ships with 0.10.0.

Issue #857, opened 2026-08-04
Local Quick Start copy connection string action implies manual connection setup is required


tnaum-ms:
Found during the `0.10.0-bug-bash-1` bug bash.

## Summary

Once DocumentDB Local starts, the wizard offers to copy the connection string. This led a user to believe they needed to create a connection manually, even though the Local Quick Start connection was already available in the Connections view.

## Steps to reproduce

1. Start DocumentDB Local through Local Quick Start.
2. Observe the post-start actions in the wizard.
3. Select or notice the option to copy the connection string.

## Expected

The user understands that the started local instance has already been added to the Connections view and is ready to use.

## Actual

The copy action suggests that manual connection creation is the next required step.

## Impact

Users can duplicate setup work or become confused about whether DocumentDB Local is actually connected.

## Suggested fix

Hide the copy connection string action from the post-start wizard, or replace it with a clearer action that opens or focuses the existing DocumentDB Local connection. Any remaining copy action should state that it is optional and intended for external clients.

tnaum-ms:
Additional bug-bash context: checked all issues created on 2026-08-04 and searched existing Local Quick Start copy-connection-string reports. #851 also mentions Copy Connection String, but it concerns an action silently doing nothing from the Command Palette. This issue is distinct: the visible post-start wizard action is misleading because the connection already exists.

tnaum-ms:
Fixed by #866, merged into `feature/local-quickstart` as 2803a385: the success screen states outright that the connection already exists, and the copy action is an optional inline step rather than a footer button that read as the next required step.

Closing by hand rather than automatically: #866 targeted `feature/local-quickstart`, not the default branch, so GitHub ignores closing keywords and creates no link. Ships with 0.10.0.

Issue #856, opened 2026-08-04
Local Quick Start links Windows Docker setup to the wrong download path


tnaum-ms:
Found during the `0.10.0-bug-bash-1` bug bash.

## Summary

On Windows, the Local Quick Start Docker setup link should direct users to Docker Desktop. The current guidance sends Windows users to an unsuitable Docker installation path.

## Steps to reproduce

1. On Windows, open the DocumentDB Local Quick Start experience.
2. Follow the Docker setup link shown when Docker is not available.

## Expected

The Windows-specific setup link opens the Docker Desktop download or installation page.

## Actual

The link does not direct the user to Docker Desktop, making the required Windows setup less clear.

## Impact

Docker Desktop is the standard supported Docker experience for this workflow on Windows. A mismatched link increases setup friction before the user can start DocumentDB Local.

## Suggested fix

Use a platform-specific Windows link that points directly to Docker Desktop, while preserving the appropriate guidance for other platforms.

tnaum-ms:
Additional bug-bash context: checked all issues created on 2026-08-04 and searched existing Docker and VS Code restart reports. No existing issue describes the missing post-install restart instruction on Windows. This is separate from #839, which concerns Docker readiness diagnosis.

tnaum-ms:
Fixed by #866, merged into `feature/local-quickstart` as 2803a385: the install CTA follows the guide the host resolved for the platform instead of a hardcoded Docker Engine URL, so Windows and macOS reach Docker Desktop.

Closing by hand rather than automatically: #866 targeted `feature/local-quickstart`, not the default branch, so GitHub ignores closing keywords and creates no link. Ships with 0.10.0.

Issue #855, opened 2026-08-04
Local Quick Start does not tell Windows users to restart VS Code after installing Docker Desktop


tnaum-ms:
Found during the `0.10.0-bug-bash-1` bug bash.

## Summary

After installing Docker Desktop on Windows, the Local Quick Start flow does not explain that VS Code must be restarted before the extension can detect and use Docker.

## Steps to reproduce

1. On Windows, open DocumentDB Local Quick Start without Docker Desktop installed.
2. Install Docker Desktop from the setup guidance.
3. Return to the already-running VS Code window and retry the Local Quick Start flow.

## Expected

The workflow clearly tells the user to restart VS Code after Docker Desktop installation, or refreshes its Docker environment reliably without a restart.

## Actual

There is no restart instruction. Users can remain in a state where Docker is not detected and do not know that restarting VS Code is required.

## Impact

The setup appears broken even after Docker Desktop is successfully installed.

## Suggested fix

Add a Windows-specific post-install instruction that explicitly asks the user to restart VS Code before retrying. Consider a retry or refresh action after restart as part of the flow.

tnaum-ms:
Additional bug-bash context: checked all issues created on 2026-08-04 and searched existing Local Quick Start and Docker Desktop issues. No issue covers the Windows setup link destination. Related issue #839 concerns Docker daemon readiness only, not setup guidance.

tnaum-ms:
Fixed by #866, merged into `feature/local-quickstart` as 2803a385: Windows and macOS guidance says to restart VS Code after installing Docker Desktop, and states that reloading the window is not enough.

Closing by hand rather than automatically: #866 targeted `feature/local-quickstart`, not the default branch, so GitHub ignores closing keywords and creates no link. Ships with 0.10.0.

PR #866, opened 2026-08-05
Fix the Local Quick Start bug-bash issues (#851, #852, #855, #856, #857, #858)

guanzhousongmicrosoft:
Fixes the Local Quick Start issues from the `0.10.0-bug-bash-1` bug bash, plus a regression suite so the next bash does not rediscover them.

@tnaum-ms: #855 and #856 were assigned to you, so those two commits are in your area; happy to hand them back if you'd rather take them yourself.

> **Maintainer note (2026-08-06).** This branch predated 23 commits of code-review work on the base, so the base has been merged in and the overlaps resolved in place. Two things also left the PR: the #854 bind-time port work (see the comment thread; #854 is being closed as won't-fix) and the Playwright visual suite, which moved to a parked draft PR #867 with authorship preserved. Everything else is unchanged.

## Fixes

**#851, palette commands silently did nothing.** All eight `localQuickStart.*` commands were listed in the Command Palette because none had a `commandPalette` entry. Invoking a lifecycle command there found no instance and returned in silence: no notification, no error, no log line. The seven context-only commands are now gated `"when": "never"`, leaving only `.open`. As a second line of defense, any route into a lifecycle command with nothing to act on explains itself, including the credential-unavailable state the service already detected internally but never surfaced.

**#852, port errors always rendered in English.** Two user-facing errors were raw template literals that never passed through `l10n.t()`, while the success message a few lines away was localized. Both are wrapped, along with the sibling strings in the same file that reach the webview.

**#856, Windows sent to the wrong Docker.** The real defect was a layer deeper than the issue described: the "Install Docker" button **hardcoded** the Docker Engine URL and ignored the guide the host had already resolved, so a per-platform mapping alone could not have taken effect. The CTA now follows the resolved guide; Windows and macOS reach Docker Desktop, WSL and remote targets keep the Engine guide.

**#855, no restart instruction on Windows.** After installing Docker Desktop, a VS Code that was already running keeps its old PATH, so Docker stays undetected and setup looks broken. Windows and macOS guidance now says to restart VS Code, and that reloading the window is not enough (the extension host respawns from the same main process and inherits the same environment).

**#857, success screen implied setup wasn't finished.** "Copy Connection String" sat next to "Open Connection" in the footer and read as the next required step; a bug-bash user followed it and created the connection by hand. The screen now states outright that the connection already exists, and the copy action becomes an optional inline step scoped to clients outside VS Code.

**#858, duplicate localhost connections.** Adding a localhost connection by hand never checked the managed instance: it's service-owned and in-memory, so the stored-connection scan couldn't see it. The wizard now offers Open Existing / Add Anyway / cancel. Host comparison is normalized (`localhost` / `127.0.0.1` / `::1`, bracketed and bare IPv6); `127.0.0.2` stays distinct on purpose, since a service bound to one loopback address isn't reachable on another.

**Drive-by, found while testing:** the success screen's **"Open Connection" did nothing**. It ran `connectionsView.focus` and nothing else, which is invisible when that view is already active, the normal case since Quick Start is opened *from* it. It now reveals, selects, and expands the instance row, so the databases load.

## Verification

- **208 Jest suites / 3396 tests** green, `tsc` / `eslint` / `prettier` clean. That is +3 suites and +42 tests over the base branch (205 / 3354).
- The three new suites are `contributions.test.ts` (the `commandPalette` gating for #851, which lives in package.json data and so is unreachable from any service-level test), `localEndpoint.test.ts` (#858) and `revealQuickStartInstance.test.ts` (the tree paths every "take me to the instance" action depends on).

## Notes for review

- #852's scope grew slightly: I localized the adjacent strings in the same file rather than only the two named in the issue, since they'd fail the same way at the next bash.

## Issues

Fixed here: #851, #852, #855, #856, #857, #858.

**These did not auto-close on merge, and were closed by hand instead.** This PR targeted `feature/local-quickstart`, not the default branch, so GitHub ignores closing keywords there: it registers no link and merging has no effect on the issues. The `Closes #NNN` keywords that originally sat here were inert. All six were closed manually against merge commit `2803a385`, each with a comment saying what fixed it.

Worth knowing for next time: the same applies to #798, whose base is `release/0.10.0`. The only PR in this chain where closing keywords actually fire is #820, `release/0.10.0` into `main`.

Two of the six were already partly fixed on the base branch by the code-review work, and this PR closes out the remainder:

- **#851**: the `"when": "never"` palette gating landed with review finding M3. What this PR adds is the explanatory `"//"` key and the guards that make a lifecycle command with nothing to act on say so instead of returning in silence.
- **#852**: most of the localization landed with review finding M1/WP-4. What this PR adds is the one instance that pass missed, a bare `'in progress'` marker that reached the message bar untranslated, plus `contributions.test.ts` to keep those strings in the bundle.

#854 is closed as not planned; see the comment thread for the reasoning.

https://claude.ai/code/session_01UbEVtZ3XxHAm3zSkkFPaCq


tnaum-ms:
@guanzhousongmicrosoft thank you!

tnaum-ms:
Thanks for this, @guanzhousongmicrosoft. Seven issues closed and three of them turned out to be deeper than the report described, which is the useful kind of bug-bash follow-up.

One logistics note first: this branch was cut from `1365964a`, which is the tip of `feature/local-quickstart` **before** 23 commits of code-review work landed. So it predates all of it, and a plain merge would have collided in sixteen places. Rather than ask you to rebase a nine-commit branch across that, I resolved it in place on your branch. Your authorship is preserved on all seven of your commits; my resolution work sits on top as a merge commit plus two follow-ups.

**What changed, and why**

1. **`ae25ac3b` (#854) is dropped.** We are keeping the explicit-port model the review landed. Details in a separate comment; #854 is being closed as won't-fix so nobody silently re-fixes it.
2. **The Playwright suite moved to a parked draft PR, #867.** Not a rejection. Details in a separate comment.
3. **`contributions.test.ts` stays here.** It is release relevant and it is Jest, not Playwright.
4. **The base is merged in and the overlaps resolved.** Details in a separate comment.

**Where it stands**

`208 suites / 3396 tests` green, `tsc` / `eslint` / `prettier` clean. That is +3 suites and +42 tests over the base branch (`205 / 3354`), which is exactly your three new suites.

The PR is now 15 files instead of 25, and every remaining line is release relevant.

tnaum-ms:
## #854 / `ae25ac3b`: dropping it, and closing the issue

Thank you for pre-isolating this in its own commit and offering it up. That made the decision cheap to act on, and it is the reason the rest of the merge was straightforward.

We are keeping the port model the code review landed (finding L3), and the principle behind it is **"no magic after execute"**:

- the **Configure** step suggests a free port and validates it while the user can still react
- so the port is **always explicit**, and it is the one the user saw before pressing Start
- a conflict at bind time is therefore a **hard, explained error**, not a silent relocation

Your version buys real edge-case robustness. The probe-release-pull-bind window is genuinely a TOCTOU, and you are right that the old "all in use" message could not fire once the check had passed. But the explicit model closes that window from the other side: there is no probe-then-release gap to lose, because the port is fixed and validated before anything is pulled, and if it is gone by bind time the user is told which port and what to do about it. What your version adds on top of that is an automatic relocation, and that is the part we do not want. A user who chose 10260 and ends up on 61146 has to discover that from the success screen.

Concretely, the parts of `ae25ac3b` this displaces:

- `findAvailablePort` is already gone. You were right that it probed a sample of the band and then claimed the whole band; the review deleted it outright rather than rewording it.
- the bind-time retry with the host port omitted has no counterpart, by design
- `boundPort` survives and still reads the actual binding, so the success screen and the stored connection string cannot drift from reality

**#854 is being closed as won't-fix**, with this reasoning recorded on the issue, so it does not get quietly re-fixed later.

One knock-on: your `99a03325` also corrected the "all in use" wording. That string no longer exists on the base, so that hunk resolved to a no-op. Same for the assertion in `contributions.test.ts` that guarded against reintroducing it; I repointed it at the current contract instead (see the merge-resolutions comment).

tnaum-ms:
## The Playwright suite: deferred, not rejected

This is genuinely good work and I want to be precise about why it is not landing here.

**It implements our own future-work doc.** `docs/ai-and-plans/live-preview-playwright-future-work.md` lists what a committed webview harness would need. This does three of the four actionable items: a committed harness instead of the throwaway page we recreated every session, a fixture set for host responses, and **theme switching**, which that doc calls _"the largest current gap"_. It also correctly **declines** the one the doc warns is premature, committed pixel baselines, and your reasoning for declining it (OS-rendered text makes cross-machine baselines meaningless) is the same reasoning the doc gives. The mutation test, reverting the fixes and confirming exactly the matching assertions fail, is the right way to prove a suite is not testing itself.

**Why it still cannot land now.** Every layer it stands on is being replaced:

- **webpack is going away.** The harness's whole premise is `dist/views.js` produced by `webpack-dev-wv` and served as a site root.
- **the test story is being redone.** `npm test` has been a no-op stub since the legacy `vscode-test` suite was retired, with a replacement due in 0.10.0. Adding a second, separately invoked runner just before that lands fragments it further.
- **the wire protocol it hand-stubs is moving.** The harness reimplements tRPC-over-`postMessage` by hand (`{id, op:{type, path, input}}` in, `{id, result}` back). That plumbing is moving into `@microsoft/vscode-ext-webview`.

Three legs, all changing. Building the visual suite on the layer we are about to delete would mean rewriting it immediately afterwards.

**Where it went: #867**, a parked draft against `main`, with your authorship preserved on all three commits. `main` has zero Local Quick Start files, so the suite cannot run there yet; that is stated plainly in the PR body, and it is the point of parking it rather than deleting it. The body also records what I added on top:

- `devServer.devMiddleware.writeToDisk` in `webpack.config.views.js`. Without it `watch:views` builds in memory and the harness silently renders a stale bundle, which is the single biggest time-waster in that workflow.
- a note in `.github/copilot-instructions.md`, so agents know the harness exists. Being able to drive a real webview state from an agent is most of its value.
- removal of the `success-relocated-port` scenario, which depended on the #854 relocation we are not taking.

It also records the fixture drift to fix on the way back in (`willReuse` is now `canReuseExistingData`; `suggestedPort`, `checkPort` and `onInstanceChanged` are missing from the stub), and the open decisions: CI versus explicitly manual, and whether this becomes the new `npm test`.

tnaum-ms:
## Merge resolutions

Only four files conflicted, down from the sixteen hunks a naive merge produced, because dropping `ae25ac3b` took `ContainerRuntime.ts`, `quickStartTypes.ts` and `QuickStartService.test.ts` out of the PR entirely. Full reasoning is in the merge commit message; the parts worth your attention:

**1. `LocalQuickStart.tsx`, the Docker guidance maps. Neither side verbatim.**

The review converted `DOCKER_GUIDANCE` and `DOCKER_GUIDES` from module-scope consts into render-time functions (`dockerGuidance()`, `dockerGuides()`). That is mandatory, not stylistic: `WebviewRegistry` imports the component statically, so the module body runs **before** `l10n.config()` does inside `render()`, and a module-scope `l10n.t()` is permanently English. It is finding M1, tracked as #864.

Your `bc08cfd1` added four keys to those same maps. Taking your shape verbatim would have shipped your new Windows and macOS strings untranslatable; taking ours verbatim would have dropped them. So: our function form, your four keys inside it.

There was also a third reference the merge did **not** flag, because nothing conflicted textually: the #856 install CTA in the JSX still read `DOCKER_GUIDES[shownDocker.presentation.guide]`. It merged silently and would not have compiled. Repointed at `dockerGuides()[...]`.

**2. `handleInstallDocker`: took your deletion.** You are right that it is not just dead code. It hardcoded `dockerGuides().install.href` and ignored the guide the host had already resolved, and that hardcoded URL **is** defect #856. The per-platform mapping cannot take effect while it exists.

**3. The `'in progress'` marker: your version wins, and this was a real bug we missed.** In `resumeReadiness`, the base yielded `error: 'in progress'` alongside a localized `message`. The webview renders `event.error ?? event.message`, so that bare untranslated marker reached the message bar verbatim. Our own M1/WP-4 localization pass walked straight past it. Good catch.

**4. `provision()` signature: took ours,** `(signal, options, alias = DEFAULT_ALIAS)`. Review N5 threaded `alias` consistently for multi-instance work and your branch predates it. Your hoisted message const is kept, it reads better.

**5. `package.json` auto-merged cleanly but wrongly.** Both branches added the same seven `"when": "never"` `commandPalette` entries at different offsets, so git kept **all fourteen**. Deduplicated, keeping your block because it carries the explanatory `"//"` key. Net effect versus the base is zero added entries plus that comment.

Because every presence check still passed against the duplicated block, I added a **uniqueness** assertion to `contributions.test.ts`. Presence alone cannot catch that class of merge artifact.

**6. Two stale assertions in `contributions.test.ts`.** Both fallout from the port model, not from anything you did:

- it asserted `'Port {0} was busy, using {1} instead.'` reaches the bundle. That string no longer exists anywhere.
- the port-conflict wording is now _"Port {0} is already in use. Go back to Configure to pick a different port, or free it, then try again."_
- the guard against reintroducing _"ports ... are all in use"_ guarded a string that had already been deleted. I kept the guard but repointed it at the current contract: no string may tell the user the port was substituted.

**7. `revealQuickStartInstance.test.ts` spied on `QuickStartService.refreshLiveState`.** Review I2-5 moved the tree's call to `refreshLiveStateInBackground()`, so the spy stopped intercepting and the test ran the real background probe. It might still have passed, which is worse than failing. Repointed. The path constants it guards are unchanged and still match what `LocalQuickStartItem` builds, which I verified.

**8. Something that fixed itself.** Our I2-3 Configure guard added an "Open Connection" button wired to the same no-op `openConnection` handler you found, so we had widened the exact gap you were closing. Your `246e6703` repoints that procedure at `revealQuickStartInstance`, so the guard's button is fixed by construction. Nothing to do.

**9. Copy.** Three of the new strings used an em dash; house style has none in product copy. Rewritten with a colon or a sentence break, wording otherwise unchanged. `l10n/bundle.l10n.json` was regenerated at the end, never hand-merged.

tnaum-ms:
## Two small asks, both on #867 rather than here

1. **Pin `@playwright/test` to the latest stable.** It is currently a caret-ranged alpha, `^1.63.0-alpha-2026-07-29`, and the lockfile resolves both it and `playwright-core` from an ADO mirror with `sha1` integrity hashes rather than npmjs with `sha512`. A caret range over an alpha means two contributors can install different code from the same lockfile-adjacent manifest, and `sha1` is not something we want to add to the dependency graph deliberately. Neither is urgent while the PR is parked, but both should be settled before it is unparked.

2. **Decide CI versus explicitly manual, and write the decision down.** Unwired, a visual suite quietly rots into fake coverage: it passes because nobody runs it. Wired, it needs a browser image in CI and it becomes a gate on every PR. Either is defensible; leaving it undecided is not. Related: whether this becomes the "new framework" that `npm test` has been promising since it turned into a no-op stub.

Both are captured in the #867 description already, so this is really just flagging them as yours to answer when we pick it back up.

Issue #687, opened 2026-05-28
build: migrate from CommonJS to ESM


tnaum-ms:
## Context

The extension is currently CommonJS. This blocks us from consuming a growing number of modern, ESM-only dependencies. Concrete recent example: `p-limit@4+` is pure ESM, which is why #685 introduced an in-house `ConcurrencyLimiter` instead of pulling a battle-tested library off the shelf.

`vscode-cosmosdb` (sibling extension, same org) has already completed this migration on `main`, so there is a working reference for what the end state looks like.

## What this issue tracks

The actual migration. The choice of build system, test runner, linter/formatter, and Node target is **still to be made** as part of this work. Options to evaluate:

- **Build:** Vite (what vscode-cosmosdb chose) vs. continuing with webpack in ESM mode vs. esbuild vs. tsup.
- **Test runner:** Vitest (what vscode-cosmosdb chose) vs. Jest with ESM support vs. node:test.
- **Linter/formatter:** continue with ESLint + Prettier vs. switch to oxlint / oxfmt (faster, what vscode-cosmosdb chose).
- **Node engine:** bump to >=22 (matches vscode-cosmosdb) or hold lower.
- **TypeScript:** bump to latest (vscode-cosmosdb is on TS 6.x).

A short design note before the implementation PR is expected.

## Concrete benefits

- Unblocks ESM-only dependencies: `p-limit`, `chalk`, `node-fetch`, `nanoid`, `execa`, and a long and growing tail.
- Smaller bundles and faster cold start (tree-shaking is much more effective on ESM).
- Faster build/test feedback loops if we adopt Vite + Vitest (sub-second incremental builds reported by vscode-cosmosdb).
- Aligns us with vscode-cosmosdb tooling, lowering the cognitive cost of moving between repos.
- **Removes the mocha-based VS Code test runner pain point.** The legacy `@vscode/test-electron` + mocha pipeline does not love ESM; vscode-cosmosdb works around this with a `pretest` hook that drops a CJS `package.json` into `out/`. The upstream story is that this runner has either already been removed or is on track to be removed, so the workaround should be temporary or unnecessary depending on when we land.

## Known costs / risks

- Touches every file with a relative import (`./foo` -> `./foo.js` in many ESM toolchains, depending on bundler/`moduleResolution`).
- Webpack config rewrite or replacement.
- Jest -> alternative test runner migration (touches every test file's imports and mocks; `jest.mock` does not have a 1:1 ESM equivalent).
[...]

PR #880, opened 2026-08-10
[wip] stack modernization

tnaum-ms:


tnaum-ms:
Flagging #933 for this work — it is a concrete, already-shipped instance of a hazard this modernization will touch everywhere.

**Short version:** `await import('bson')` resolved the package's ESM export condition while `mongodb` resolved the CJS one. Two copies of `bson` in the bundle, two sets of classes, so every `instanceof ObjectId` check returned `false`. The schema analyzer silently classified BSON wrappers as plain objects and offered `_id.buffer` / `rating.value` in autocompletion instead of `_id` / `rating`. Six releases shipped it over 92 days before anyone noticed, because nothing ever threw.

**Why it is relevant to the stack modernization specifically:**

- **The trigger was a build-config change.** The bug appeared the day `ignoreDynamic: true` was added to the swc loader so webpack could split lazy dependencies. No product code changed. This modernization is a very large build-config change.
- **Migrating to ESM inverts the hazard rather than removing it.** Once everything is `import`, the CJS copy becomes the odd one out; any dependency still pulled in through `require` transitively reintroduces the split.
- **Nothing in our toolchain catches it.** TypeScript, ESLint and the unit tests were all green — unit tests construct values from the same copy the assertion uses, so the mismatch only exists in a bundled build.

Two things from #933 that may be worth adopting here:

1. A bundle-level guard that no module specifier resolves to two distinct files. Ours is currently a hardcoded one-liner (`grep -c vendors-node_modules_bson_lib_bson_node_mjs dist/main.js` must be 0, since webpack does not clean `dist/`); a general version would be better placed in the new build.
2. Duck-typing instead of `instanceof` at package boundaries where a discriminator exists. `BSONTypes.inferType()` now falls back to `_bsontype`, so it degrades correctly regardless of how the bundler resolves the package.

The three `import { EJSON } from 'bson'` sites now carry an inline comment explaining why the static form is load-bearing, cross-referencing #687. Full write-up with the timeline and decisions is in `docs/ai-and-plans/features/completions-and-schema/iterations/09-bson-dual-package-hazard.md`.

PR #929, opened 2026-09-17
Reduce noisy tree trace output

tnaum-ms:
## What changed
Tree and storage diagnostics now emit count-based summaries per load or cache batch rather than a trace message for every storage wrapper or constructed cluster model.

## Why
Expanding the Connections, Azure Resources, and discovery trees produced a large volume of per-item trace output that obscured useful diagnostics. The summaries retain item-type and failure counts while removing high-volume identifiers and names.

## Scope
This PR does not change tree rendering, storage behavior, telemetry, or detailed warnings for corrupt stored items and lookup failures.

## Validation
- Added a focused storage trace-summary test.
- `git diff --check` passed.
- Jest and the build could not run locally: Node/npm were absent, and dependency restoration was blocked by a missing package in the configured proxy and a TLS handshake failure against the public registry. CI should provide the remaining validation.

## Links
- Feature context: `docs/ai-and-plans/features/connections-tree/README.md`


tnaum-ms (commented):
👍 Approved.

languy (approved):

PR #951, opened 2026-09-22
local quickstart: keep credentials out of the output channel and temp dir (issue #947)

### LINKED ISSUE #947: Local Quick Start: generated password is written in plaintext to the "DocumentDB Local Setup" output channel
Found testing v0.10.2 (`c39c9b0b`). `src/services/localQuickStart/` is unchanged on `main` (`fa7bcbf5`), so the line numbers below are for `main`. Image `latest` = 0.117.0, on WSL2 with Docker Desktop 29.8.

| ID | Sev | Bug |
|---|---|---|
| SEC-1 | P1 | Password printed in plaintext on every setup, via `docker container inspect` output |
| SEC-2 | P3 | Plaintext `.env` with the credentials stays in the temp dir after an extension-host crash, for up to 1 h |
| SEC-3 | P3 | A 1-character password makes the masker replace every occurrence of that character: 325 of 406 lines turned to `***` |
| SEC-4 | P3 | The full `docker info` JSON is dumped into the channel (noise) |

### SEC-1: password in the output channel

1. Set up with defaults (auto-generated credentials).
2. Copy the connection string.
3. Output → "DocumentDB Local Setup": search for the password. It appears in the `docker container inspect --format '{{json .}}'` line, inside `"Config":{"Env":["USERNAME=…","PASSWORD=…"]}`.

- Reproduced on every setup in the run, including custom credentials.
- **Where:** `ContainerRuntime.ts:240, 249, 254, 259, 265, 317` build their runner with `makeRunner([])`, so there's nothing to mask. The `run`/`exec` paths do pass secrets. By code reading, start, stop, refresh and reconcile also inspect, so they leak too. That hasn't been run yet.
- **Expected:** no password in the channel from any command. Either don't echo `inspect` stdout, or have every runner mask the alias's known secrets.
- **Done when:** a test that captures channel output across provision → stop → start → refresh → reconcile asserts the password never appears.

### SEC-2: env file survives a crash

1. Start setup.
2. During "Creating container", kill the extension host (`kill -9`).
3. Check `ls -l /tmp/documentdb-quickstart-*.env`.

- **Result:** the file is still there (mode 0600) with `USERNAME=`/`PASSWORD=`. It's written at Creating and removed only in `provision()`'s `finally` (`QuickStartService.ts:976`), so it's on disk for the whole Creating → Waiting → sample-data window, up to 3+ minutes. `sweepStaleQuickStartEnvFiles` (`QuickStartService.ts:2185-2194`) only removes files older than 1 h.
- **Expected:** swept on the next activation. For example, put the owning PID in the file name and remove files whose PID is dead, so another window's in-flight run is safe.

### SEC-3: short password turns the log to `***`

- Custom password `e` → the channel is unreadable.
- **Expected:** a minimum password length in the router and the webview (8 characters is conventional), or masking that doesn't apply to trivially short secrets.

### SEC-4: `docker info` JSON in the channel

- Hundreds of lines of noise on every readiness check. Log a summary line (server version, OS type, architecture) instead.

### Testing

Masking is unit-tested in `outputMasking.test.ts`. For SEC-1, capture the channel through a fake runtime (see `mockRuntime` in `QuickStartService.test.ts`) whose `inspectContainer` output includes `Config.Env`, and assert the password never reaches the channel.

guanzhousongmicrosoft:
Fixes #947.

Every Quick Start setup, and every stop/start/refresh/reconcile after it, prints the instance password to the "DocumentDB Local Setup" channel. The runner echoes all stdout, and `docker container inspect` output includes the container env (`PASSWORD=…`).

Masking isn't enough, because the lifecycle callers don't always know the password (adopted containers, a missing stored secret). So `inspect` no longer echoes its stdout.

The temp `--env-file` now lasts only as long as `docker run`; Docker copies its values into the container config, so nothing reads it again. Its name carries the owning PID, so the activation sweep removes a crashed host's file right away while keeping another window's in-flight one.

SEC-3 is left as is on purpose: every password is masked, however short. A trivially short one makes the log hard to read, which beats printing it. `docker info` now logs a one-line summary instead of its full JSON (SEC-4).

Also fixes the forced flush on long newline-less output, which could cut a password in half and print both halves unmasked.

The SEC-1 test lives at the `ContainerRuntime` layer because the service tests mock the runtime, and that's where the leak was.


tnaum-ms (approved):
Code review approval for 7ddb0703b01e1bbc6dcb47bb0026abb37018a658. No blocking high- or medium-severity issue found in the reviewed changes. The focused Quick Start regression suites passed: 335 tests across 16 suites. This code approval does not replace the committed contributor AI pre-review record required by CONTRIBUTING.md.

PR #953, opened 2026-09-23
local quickstart: reject credentials and ports setup can't honor (issue #949)

### LINKED ISSUE #949: Local Quick Start: wizard accepts credentials and ports that setup can't honor
Found testing v0.10.2 (`c39c9b0b`). Service and router are unchanged on `main` (`fa7bcbf5`). The webview was refactored on `main`, and INPUT-6/7 are re-checked there; line numbers are for `main`. Image `latest` = 0.117.0.

Validation lives in two places that must agree: the router zod schema (`localQuickStartRouter.ts:55-80`) and the webview (`LocalQuickStart.tsx:899-950`).

| ID | Sev | Input | What happens |
|---|---|---|---|
| INPUT-1 | P2 | Username of 64–128 characters (UI allows 128) | 3-minute timeout. PostgreSQL role names max out at 63 **bytes**, so multibyte names hit the limit sooner. |
| INPUT-2 | P2 | Password that fails SASLprep, e.g. `pass😀word` | 3-minute timeout. The driver rejects it client-side: `Unassigned code point`. |
| INPUT-3 | P2 | Username `documentdb` | The container exits within seconds, followed by a 3-minute timeout |
| INPUT-4 | P2 | Username with a space, e.g. `my user` | Reports **Running** after 133 s, but every write fails. The gateway logs `missing "=" after "user" in connection info string`. Sample data is silently skipped. |
| INPUT-5 | P3 | Password `  abc  ` | Silently trimmed to `abc`. U+2028 and U+0085 are accepted. |
| INPUT-6 | P2 | Hidden custom-credential values | Still validated after switching back to auto-generate, so Start stays disabled with no visible field to fix |
| INPUT-7 | P2 | Empty port field | The UI shows `localhost:<suggestedPort>`, but setup binds 10260 |

### INPUT-1 to INPUT-5 (repro: Configure → custom credentials → enter the value → Start)

**Expected**, enforced in both the router and the webview:
- Username: at most 63 UTF-8 bytes, no whitespace, not `documentdb`.
- Password: must pass SASLprep. `saslprep` already ships with the `mongodb` driver.
- Reject leading/trailing whitespace instead of trimming it, and reject Unicode line/paragraph separators.

### INPUT-6: hidden fields block Start (webview)

1. Configure → Use custom credentials.
2. Enter a 129-character username.
3. Switch back to auto-generate.

- **Result:** Start stays disabled, and the error points at a field that is no longer shown.
- **Where:** `LocalQuickStart.tsx:929-940`. The length and control-character checks aren't gated on `useCustomCredentials`, though the both-or-neither check at `:923` is.
- **Expected:** skip every credential check when `!useCustomCredentials`.

### INPUT-7: an empty port shows one port and binds another (webview)

1. Hold 10260 (`python3 -m http.server 10260 --bind 127.0.0.1`), so the suggested port becomes something else.
2. Open Configure and clear the port field.

- **Result:** the summary shows the suggested port (`effectivePort`, `:1571`), but `opts.port` is only set when the field is non-empty (`:965`). The service then falls back to `QUICK_START_PORT` (`QuickStartService.ts:638`), and setup fails with "port 10260 in use".
- **Expected:** send `effectivePort`. The comment at `:961-964` already says the port is always sent.

**Done when:** router unit tests reject each INPUT-1 to INPUT-5 value, and webview tests cover INPUT-6 and INPUT-7.

### Testing

Router validation: `src/webviews/documentdb/localQuickStart/localQuickStartRouter.test.ts`. The webview checks live in `LocalQuickStart.tsx` (`advValidation`, and the `useEffect` that builds `opts`). Consider moving them into a pure function both the router and the webview share, so the two can't drift.

guanzhousongmicrosoft:
Fixes #949.

Custom credentials could pass the Configure step and then fail inside DocumentDB Local: a 3-minute readiness timeout, a container that exits at once, or a "Running" instance whose writes all fail. An empty port field showed the suggested port, but setup bound 10260.

The router and the webview now share one rule set, so they can't drift. Beyond the issue's cases, the rules come from running image 0.117.0 against about 80 usernames and passwords, e.g. `\`, `=` or `,` in a username, reserved `pg`/`citus`/`documentdb` prefixes, or a backslash in a password, which the image splices into JSON.

SASLprep is the one check the webview can't run: the driver's tables need `Buffer` and add about 560 KB to the bundle. The host checks non-ASCII passwords as the user types, and again at Start.

Credentials are no longer trimmed. Surrounding whitespace is rejected instead, so the stored password is the one the user typed.


tnaum-ms (approved):
Approved after clarifying the username hint in 6fb9b31f: up to 63 ASCII characters, fewer for non-ASCII text. The localization bundle was regenerated, prettier-fix and the build passed, and all 23 Quick Start suites / 457 tests passed. This follow-up changes only the hint and its localization entry. A separate eligible reviewer may still be required because this account pushed the follow-up; the earlier documented review-record and manual-testing caveats remain.

tnaum-ms:
approved

tnaum-ms (dismissed):

GuanzhouSong (approved):

tnaum-ms (approved):

PR #954, opened 2026-09-23
local quickstart: explain setup failures instead of exit codes (issue #948)

### LINKED ISSUE #948: Local Quick Start: setup failures show "Process exited with code N" or a 3-minute "still initializing"
Found testing v0.10.2 (`c39c9b0b`). `src/services/localQuickStart/` is unchanged on `main` (`fa7bcbf5`). Image `latest` = 0.117.0, on WSL2 with Docker Desktop 29.8. Related: #840, #854.

| ID | Sev | Bug |
|---|---|---|
| ERR-1 | P2 | Every Docker CLI failure during pull/create is reported as `Process exited with code N`. `isPortAllocationFailure` never matches. |
| ERR-2 | P2 | Readiness keeps retrying for 180 s when the container is dead or auth is rejected, then says "still initializing" and offers Wait longer |
| ERR-3 | P3 | No deadline on the creating stage. Cancel leaves the `docker container run` CLI process orphaned. |
| ERR-4 | P2 | Docker CLI not installed (Linux/WSL) is reported as "Docker daemon unreachable" |
| ERR-5 | P3 | A temp dir containing `'` breaks `docker run`: `Process exited with code 2` |

### ERR-1: Docker's stderr never reaches the user

| Repro | Shown | Docker actually said |
|---|---|---|
| Advanced → image tag `0.117.0-nope` | `Process exited with code 1` | manifest unknown |
| `docker run -d --name vscode-documentdb-local alpine sleep 1d` (unlabelled), then Set up | `Process exited with code 125` | container name already in use |
| `docker rmi` the image so the pull is slow; during Pulling run `python3 -m http.server 10260 --bind 127.0.0.1` | `Process exited with code 125` (`unexpectedFailure`) | port is already allocated |
| `docker pause vscode-documentdb-local`, then Start in the tree (the tree shows it as Stopped) | `Process exited with code 1` | container is paused; unpause it |

- **Where:** the pull, create and start paths reject with the bare `ChildProcessError` (#840 item 1). `isPortAllocationFailure` (`QuickStartService.ts:103`, used at `:895`) matches on `error.message`, which never contains stderr, so it is dead code.
- **Expected:** capture stderr for pull and create the way `dockerProbes.ts` `CapturingTeeWritable` already does for probes. Then the port case maps to `portInUse`, and the others show Docker's first error line.

### ERR-2: a dead container or rejected credentials wait out the full 180 s

`waitForReadiness` (`QuickStartService.ts:1219-1245`) retries every error until the deadline and never looks at the container.

| Repro (custom credentials) | Probe error | Result after 184 s |
|---|---|---|
| username `documentdb` | `ECONNREFUSED`: the container exits within seconds | readinessTimeout + Wait longer |
| password with an emoji | `Unassigned code point` (SASLprep) | same |
| username of 64 characters | `Invalid account: User details not found` | same |

- It also happens with defaults: run `docker kill vscode-documentdb-local` during "Waiting for DocumentDB".
- **Expected:**
  - Inspect the container between attempts. If it has exited, fail at once with the exit code and the last log lines.
  - Treat auth and SASLprep errors as terminal.
  - Offer Wait longer only when the container is running and the error is still connection-level.
- The input validation that should stop these values reaching Docker is tracked in #949.

### ERR-3: a hung `docker run` hangs setup at "Creating container"

- **Repro:** Configure → port 65535 → Set up. On Docker Desktop, `docker run` never returns for this port.
- **Result:**
  - Setup sits at "Creating container" indefinitely (about 9 minutes observed).
  - Cancel returns in about 6 s and removes the container, but the `docker container run` child process is left running.
- **Expected:** a create-stage deadline with a clear message, and on cancel, kill the child if it doesn't exit.

### ERR-4: missing Docker CLI is reported as "daemon unreachable"

- **Repro:** start VS Code with a `PATH` that has no `docker`, then Set up.
- **Result:** `dockerDaemonUnreachable`. The readiness result has `cliInstalled: true, daemonReachable: false, failureKind: unknown`.
- **Where:** `DockerReadinessService.ts:447`: `cliInstalled = versionOk || infoProbe.spawnErrorCode !== 'ENOENT'`. The probes run through a shell, so a missing binary is the shell's exit 127, never `ENOENT`.
- **Expected:** treat the shell's exit 127 as "CLI missing", so the user gets `dockerCliMissing` and the install guidance.

### ERR-5: apostrophe in the temp dir breaks setup

- **Repro:** start VS Code with `TMPDIR="/tmp/qs it's"`, then Set up.
- **Result:** fails at Creating with `Process exited with code 2`. Spaces, `$` and backticks in the path work.
- **Where:** the path is passed strong-quoted through the shell runner, and `Bash.quote()` in `@microsoft/vscode-processutils` renders `/tmp/qs it's/x.env` as `'/tmp/qs it\'s/x.env'`. Bash doesn't allow escapes inside single quotes, so the shell hits a syntax error (exit 2). The env-file path is built in `ContainerRuntime.ts` (`envFiles`, `:98`).
- **Fix options:** escape `'` as `'\''` before handing the path over, or write the env file under a directory whose path we control.

### Testing

`QuickStartServiceImpl` takes its container runtime in the constructor. `QuickStartService.test.ts` (`mockRuntime`) and `QuickStartProvisionDurability.test.ts` (`runtimeFor`, with `mongodb` mocked) show how to fake Docker and drive `provision()` and the lifecycle methods without a daemon. Readiness classification has its own tests in `DockerReadinessService.test.ts` and `dockerReadinessClassification.test.ts` (ERR-4).

guanzhousongmicrosoft:
Fixes #948.

Every Docker failure during Local Quick Start setup reads "Process exited with code N". The container-client runner rejects with a bare `ChildProcessError` and drops stderr, so a bad image tag, a name clash, a port taken mid-pull and starting a paused container all look the same, and the port-in-use mapping never fires. Readiness retries for 3 minutes after the container has exited or the credentials were refused, then offers "Wait longer". A hung `docker run` (Docker Desktop, port 65535) sits at "Creating container" forever, and Cancel leaves the CLI running. A missing `docker` shows as "daemon unreachable" (the shell exits 127, never ENOENT), and a temp dir containing `'` breaks `--env-file` because `Bash.quote()` backslash-escapes inside single quotes.

Catching the error isn't enough: the library runner exposes neither stderr nor the child's pid, and `docker run` ignores the SIGTERM its tree-kill sends. Commands whose output or failure matters now go through a small runner that keeps stderr, takes a deadline, and escalates to SIGKILL on the process group.

If inspecting the container fails mid-wait (Docker gone, container removed), setup keeps waiting as today. Failing fast there means deciding what to do with a container and saved credentials we can't reach.

New surface: a failed Start/Stop/Restart/Delete, or a container that exits right after Start/Restart, now raises an error notification. Before, the tree row just turned to Error.

The Windows `cmd.exe`/`taskkill` path has no automated coverage (CI is Ubuntu only).


tnaum-ms:
Pushed 21f018ff to align the failure messages with the Quick Start wizard. Sign-in failures now use plain, neutral language; saved-credential failures name the exact Configure option and explicitly warn about permanent data loss; container and lifecycle failures point to the setup log while retaining the diagnostic detail. Existing image-tag and WSL guidance was already clear and remains unchanged. Full local verification passed: npm run l10n, npm run prettier-fix, npm run lint, npx jest --no-coverage --runInBand (276 suites, 4,257 tests, 4 snapshots), npm run build, and npm run package. This follow-up does not change subprocess behavior. Native Windows execution and real-Docker manual scenarios were not tested here. The contributor AI pre-review record required by CONTRIBUTING.md remains separate and outstanding.

tnaum-ms (approved):
Approved after the wording and localization follow-up in 21f018ff. Full local verification passed: localization, Prettier, lint, all 4,257 tests and 4 snapshots, build, and VSIX packaging. All current CI checks, including integration tests and packaging, are green. Native Windows execution and real-Docker manual scenarios were not tested in this follow-up; subprocess behavior was not changed by this commit. The contributor AI pre-review record required by CONTRIBUTING.md remains a separate outstanding process item.

tnaum-ms (dismissed):
Reaffirming approval for 21f018ff. The PR head has not changed since the previous review, and all current CI checks pass. The earlier full local verification remains applicable. Native Windows testing and the contributor AI pre-review record remain the documented caveats. This branch is behind main, and repository rules may still require a separate eligible reviewer because this account pushed the wording follow-up.

tnaum-ms (approved):

PR #958, opened 2026-09-23
local quickstart: never wipe data before setup can succeed (issue #946)

### LINKED ISSUE #946: Local Quick Start: setup deletes data volumes it didn't create, and wipes before it knows it can succeed
Found testing v0.10.2 (`c39c9b0b`). `src/services/localQuickStart/` is unchanged on `main` (`fa7bcbf5`), so the line numbers below are for `main`. Image `latest` = DocumentDB Local 0.117.0, on WSL2 with Docker Desktop 29.8.

Root cause for DATA-1 and DATA-2: the pre-clean in `provision()` (`QuickStartService.ts:716-727`) runs **before** the port check (`:731`) and the image pull. Any failure after it leaves the user with nothing.

| ID | Sev | Bug |
|---|---|---|
| DATA-1 | P1 | A normal **Set up** force-deletes an existing `vscode-documentdb-local-data` volume, even when setup then fails |
| DATA-2 | P2 | **Start fresh** deletes the container and volume, then fails on a taken port |
| DATA-3 | P3 | A volume held by a foreign container is silently shared instead of removed |
| DATA-4 | P3 | "Already seeded" checks `sampledb`, but 0.117 seeds `StoreData`, so every recreate re-seeds and restores deleted documents |

### DATA-1: normal setup deletes a volume it has no record of

```sh
docker volume create vscode-documentdb-local-data
docker run --rm -v vscode-documentdb-local-data:/d alpine sh -c 'echo keep > /d/marker'
# VS Code: DocumentDB Local → Set up, defaults (Start fresh NOT chosen)
docker run --rm -v vscode-documentdb-local-data:/d alpine cat /d/marker   # → No such file
```

- The output channel shows `$ docker volume rm --force vscode-documentdb-local-data`, with no prompt or warning.
- Repeat with image tag `0.117.0-nope`: the pull fails ("Process exited with code 1"), and the volume is **already gone**.
- A volume with any other name survives the same run, so the volume is deleted because of its name.
- **Real-world trigger:** the container is removed (`docker container prune`, Docker Desktop cleanup) while the volume stays, and this VS Code profile has no `ready` record. That covers a second profile, Insiders vs Stable, a reinstall, or lost globalState.
- **Where:** `:723-727`. The wipe is gated on "no managed container AND no `ready` record" (comment at `:703-708`), which assumes such a volume is a safe no-op. It isn't.
- **Expected:** never delete a volume this run didn't create. If the volume exists with no credentials, take the same path as `credentialsUnavailable` and offer Start fresh.

### DATA-2: Start fresh wipes, then fails the port check

1. Set up on port 10263, insert a document, then Stop.
2. Hold the port: `python3 -m http.server 10263 --bind 127.0.0.1`.
3. Run setup with `{ port: 10263, startFresh: true }`. In the UI this is a race between Configure's port check and Start.

- **Result:** `portInUse`, and the container and volume are both gone.
- **Expected:** run every check that can fail (port, tag format, Docker readiness) before any destructive step. Ideally delete the old volume only after the new image has pulled.

### DATA-3: our volume is mounted by a foreign container

```sh
docker volume create vscode-documentdb-local-data
docker run -d --name holder -v vscode-documentdb-local-data:/data alpine sleep 1d
# Set up → succeeds; holder and our container now share the volume
```

- The `removeVolume` failure is swallowed (`.catch(() => meterQuickStartSilentCatch('provision_removeStaleVolume'))`).
- **Expected:** detect that the volume is in use and stop with a clear message.

### DATA-4: seed marker is the wrong database for 0.117

- `SAMPLE_DATA_DB = 'sampledb'` (`:127`), but 0.117's `init_documentdb_data.sh` creates `StoreData` (`stores`, 41,505 docs; `ratings`). The check never matches.
- So every recreate re-runs the seed (~15 s). Duplicates are skipped, but `StoreData` documents the user deleted come back.
- **Expected:** detect by what the image actually seeded, or record "seeded" on the instance record.

**Done when:** a unit test for each of DATA-1 and DATA-2 fails on today's code (fake runtime: pre-existing volume with no record; startFresh with the port taken), and passes after the fix.

### Testing

`QuickStartServiceImpl` takes its container runtime in the constructor. `QuickStartService.test.ts` (`mockRuntime`) and `QuickStartProvisionDurability.test.ts` (`runtimeFor`, with `mongodb` mocked) show how to fake Docker and drive `provision()` and the lifecycle methods without a daemon.

guanzhousongmicrosoft:
Fixes #946.

Setup removed the `vscode-documentdb-local-data` volume before it checked the port or pulled the image. So a plain **Set up** could erase data it had no record of (container pruned, another VS Code profile, a reinstall), and **Start fresh** could erase everything and then fail on a busy port or a bad tag.

Now the volume is removed only right before `docker run`, once the port is checked and the image is local. A volume this profile has no credentials for goes through the existing credentials-missing gate, so only an explicit Start fresh can remove it. A volume that shows up while the image downloads is left alone.

- A failed volume removal now stops setup. Docker refuses to remove a volume another container mounts, so setup no longer silently shares it (DATA-3).
- A run that fails after `docker run` removes the volume it created. Otherwise the retry would hit the new gate.
- Sample data is seeded only into a volume that didn't exist before this run. The old check looked for `sampledb`, but 0.117 seeds `StoreData`, so every recreate restored documents the user had deleted (DATA-4).
- One UI change: after the gate trips, the failed screen's button reads **Start fresh** with Configure's erase warning. The button used to say "Retry setup" but would already have run a Start fresh. The strings are existing ones.

The container is still removed before the port check. That loses no data, and the port check relies on our own container being gone.

The new tests for DATA-1, DATA-2 and DATA-4 fail on `main`. End to end against real Docker, a leftover volume, a busy port, a bad tag, a held volume and a volume created mid-pull each left the existing data intact. A Cancel after `docker run` removed only the volume that run had created, and recreates seeded samples only into a new volume.

Supersedes #952 with a smaller change.


tnaum-ms:
Pushed 8527c82b to make the two volume-protection errors friendlier and localizable. The messages explain why setup stopped, point back to Configure or checking volume usage, and no longer suggest deleting another container. Added regression assertions and corrected the sample-seeding documentation. Full local verification passed: npm run l10n, npm run prettier-fix, npm run lint, npx jest --no-coverage --runInBand (275 suites, 4,218 tests, 4 snapshots), npm run build, and npm run package. No real-Docker manual testing was performed in this follow-up. The contributor AI pre-review record required by CONTRIBUTING.md remains separate and outstanding.

tnaum-ms (approved):
Approved after the wording and localization follow-up in 8527c82b. Full local verification passed: localization, Prettier, lint, all 4,218 tests and 4 snapshots, build, and VSIX packaging. All current CI checks are green. The contributor AI pre-review record required by CONTRIBUTING.md remains a separate outstanding process item.

tnaum-ms (dismissed):
Re-reviewed the current head 7e565cbc against the previously approved 8527c82b. The only subsequent commit merges main; the reviewed Quick Start implementation, localized wording, tests, and feature documentation are unchanged. All current CI checks pass. Reaffirming code approval; the previously documented contributor review-record caveat still applies. Repository rules may still require a separate eligible reviewer because this account pushed the wording follow-up.

guanzhousongmicrosoft:
09502369 — keep volumes setup can't prove it created.

- Failure cleanup removes the data volume only when this run got a container (`docker run` returned an ID, or the operation-labelled sweep found one). Before, a failed `docker run` that created nothing could remove a volume another process made after the last check.
- `volumeExists` parses strictly and stops echoing every host volume name into the output channel. Non-strict parsing drops rows it can't read, which would read as "no volume" and skip the #946 gate.
- The cleanup test now asserts the removal follows the failed create, plus a new test for a failed create with no container.

Quick Start service and webview tests: 20 suites, 363 passed. Not re-run against real Docker.

guanzhousongmicrosoft:
36fb9b2d — iteration record and review for this PR (CONTRIBUTING §6), in `docs/ai-and-plans/features/local-quickstart/iterations/06-data-volume-protection/`.

`iteration.md` covers why the PR exists, why it replaced #952, each work item with its commit and rejected alternatives, and what was and wasn't verified. `review.md` lists every review finding with its verdict and decision. Two residual risks are accepted and written down there: a volume created by another process in the millisecond before `docker run`, and Start fresh against a foreign container holding our name.

tnaum-ms (approved):

